Issue #74, opened 2026-05-20
Merge plugin buckets into a single weight-ordered, first-match-wins evaluation pass


sean-e-dietrich:
## Problem

Today `Firewall::evaluate()` runs three separate `PluginManager` instances in a fixed sequence — bypass (`response: allow`) → challenge (`response: challenge`) → block (`response: block`). Weights only sort *within* each bucket. The buckets themselves run in hard-coded order, so allow always beats block regardless of weight.

That ordering is fine in most cases but it cannot express "block this clearly malicious traffic first, *then* let any survivors through the allow list" — a common ask when you want a high-confidence block rule to outrank a broad-but-trusted allow list.

## Proposal

Collapse the three buckets into a single ordered list and evaluate it with **first-match-wins** semantics.

### Evaluation order

```
Request
  ├─ Storage blocklist check         → exit if blocked
  ├─ Challenge submission?           → validate + mint/reject, exit
  └─ Single weight-ordered plugin loop
       For each plugin in ascending weight order:
         if !matches(plugin): continue
         switch plugin.response:
           allow     → return; request allowed
           block     → block; exit
           challenge → if valid pass token: continue
                       else: serve interstitial; exit
```

### Key semantics

- **Single global weight ordering.** All plugins compete in one ascending-weight list. There is no concept of "all allows run before all blocks" anymore — only weight.
- **First match wins, period.** The matching plugin's `response:` is final. No accumulation, no "tentative" outcomes. Operators control precedence purely by choosing weights.
- **Pass tokens skip challenge plugins.** A valid pass token causes a matching `response: challenge` plugin to be treated as "no match" so the loop moves past it. Block plugins still apply — a token attests "I am human", not "I am exempt from blocks".
- **Storage blocklist still runs first** (current behavior, preserved). It is not a plugin; it is a durable "you already misbehaved" check.
- **Challenge submission interception still runs before the loop** (current behavior, preserved). The POST to `/_firewall/challenge` is special-cased and bypasses plugin evaluation, with only the storage blocklist applied.

### Worked example

```yaml
plugins:
  - plugin: Crs                  # high-confidence WAF block
    response: block
    weight: -100

  - plugin: IpAddress            # operator allowlist
    response: allow
    weight: 0
    config: [10.0.0.0/8]

  - plugin: GeoLocation          # geo-block fallback
    response: block
    weight: 100
    config: [country:NK]
```

| Request                                | Today's behavior                                | Proposed behavior                                  |
[...]

PR #91, opened 2026-07-27
Fix inverted Crs::evaluate() return value (Fixes #89)

### LINKED ISSUE #89: Crs::evaluate() return value is inverted — CRS blocks all legitimate traffic and allows real attacks
## Summary

`Crs::evaluate()` returns the opposite boolean to every other plugin in the library. The two `return` statements in `src/Plugins/Crs.php` are swapped, so the firewall **blocks benign traffic and lets genuine attacks through**.

Severity: critical. Any deployment that enables the CRS plugin with `response: block` is worse off than having no firewall at all — legitimate visitors get a 403 while SQL injection reaches the application.

This is not new in v2.8.0. `src/Plugins/Crs.php` is byte-identical to the `2.x` branch, so the defect has been present since the plugin landed in #9. v2.8.0 is simply the first release where the plugin is reachable by consumers.

## The defect

`src/Plugins/Crs.php`, `Crs::evaluate()` (declared line 66):

```php
if ($crsVerdict->isBlocked()) {
    $this->getLogger()->info('CRS blocked request', $this->getContext($request, [...]));
    return false;        // line 79 — reports "no match" when CRS decided to BLOCK
}

if ($crsVerdict->matchedRules !== []) {
    $this->getLogger()->debug('CRS matched but did not block', ...);
}

return true;             // line 90 — reports "match" when CRS decided to ALLOW
```

The logging on line 72 is correct. Only the two return values are wrong.

## Why this is the wrong polarity

`PluginManager::evaluate()` (lines 198–207) is the only consumer of the return value, and it treats truthy as *matched*:

```php
$status = $plugin->evaluate($request);
// ...
if ($status) {
    $this->getLogger()->debug('Plugin evaluation matched', ...);
```

A polarity audit across the shipped plugins confirms CRS is the sole outlier. Each was given one input that should match and one that should not:

```
                    matched-input   unmatched-input
IpAddress           true            false            consistent
Url                 true            false            consistent
UserAgent           true            false            consistent
VulnerabilityScore  true            false            consistent
Crs (mode: block)   false           true             *** INVERTED ***
```

## Reproduction

Self-contained; no framework required. Run from any project with `kanopi/firewall` v2.8.0 installed:

```php
<?php
require __DIR__ . '/vendor/autoload.php';

use Kanopi\Firewall\Plugins\Crs;
use Kanopi\Firewall\Plugins\Url;
use Symfony\Component\HttpFoundation\Request;

$sv     = ['REMOTE_ADDR' => '203.0.113.10', 'HTTP_USER_AGENT' => 'Mozilla/5.0'];
$benign = Request::create('/', 'GET', [], [], [], $sv);
$sqli   = Request::create("/?id=1' UNION SELECT 1,2,3--", 'GET', [], [], [], $sv);

// Reference polarity from a plugin that behaves correctly.
$url = new Url([], [['variable' => 'path', 'operator' => 'equals', 'value' => '/admin', 'negate' => false]]);
var_dump($url->evaluate(Request::create('/admin')));  // true  — matched
var_dump($url->evaluate($benign));                    // false — no match

$crs = new Crs([], ['paranoia' => 1, 'mode' => 'block']);
var_dump($crs->evaluate($benign));  // true  — WRONG. engine: action=allow, score=0
var_dump($crs->evaluate($sqli));    // false — WRONG. engine: action=block, score=3
```

Observed output:

```
kanopi/firewall v2.8.0

  Url        matches /admin  => true   expected true
  Url        no match        => false  expected false

  Crs        benign GET /    => true   expected false   engine: action=allow score=0
  Crs        SQL injection   => false  expected true    engine: action=block score=3
```

Note that `getLastVerdict()` is correct in both cases — `action=allow` for the benign request, `action=block` with a non-zero score for the injection. The engine is fine; only the plugin's translation of that verdict into a bool is wrong.

### Confirmed end to end over HTTP

On a Drupal 11 site with a single CRS rule at `paranoia: 1`, `mode: block`:

| Request | Result |
|---|---|
| `GET /` | **403 blocked** |
| `GET /?id=1' UNION SELECT 1,2,3--` | **200 allowed** |
| `GET /.env` (blocked by a `Url` rule instead) | 403 — unaffected plugins still correct |

The stored block record confirms attribution: `{"value":{"plugin":"CRS",...,"request":{"path":"/"}}}` — CRS blocked the homepage.

## Root cause: the interface docblock contradicts the implementation

`src/Plugins/PluginInterface.php` line 38:

```php
/**
 * Evaluate the current request to see if it passes the current plugin.
 *
 * @return bool
 *   Return TRUE if allowed to pass.
 */
public function evaluate(Request $request): bool;
```

Read literally, CRS is the **only** plugin that obeys that docblock. But it contradicts `PluginManager`, the other seven plugins, and the README's own custom-plugin example:

```php
return true; // Block request
return false; // Allow request
```

So there are really two defects: the docblock is wrong, and the CRS plugin was written against it. Fixing only the plugin leaves the same trap for the next contributor who implements `PluginInterface`.

## Why CI did not catch it

`tests/Unit/Plugins/CrsTest.php` encodes the same misreading, so the suite is green:

```php
public function testBenignRequestPasses(): void
{
    $plugin = new Crs([], ['paranoia' => 1]);
    $request = $this->request(['q' => 'hello world']);
    $this->assertTrue($plugin->evaluate($request));          // asserts TRUE for benign
    $this->assertFalse($plugin->getLastVerdict()->isBlocked());
}

public function testSqliRequestBlocks(): void
{
    // ...
    $this->assertFalse($plugin->evaluate($request));         // asserts FALSE for SQLi
    $this->assertTrue($verdict->isBlocked());
}
```

Compare `tests/Unit/Plugins/UrlTest.php` around line 237, where the comment sits directly above the opposite assertion:

```php
// Should match: 1-2 letter filenames
$this->assertTrue($plugin->evaluate(Request::create('/a.php')));
```

The two files assert opposite conventions. The CRS assertions need inverting as part of the fix, or the next change will re-break it.

The CRS test file's own docblock states it covers "the bool return contract of evaluate()" — it does, just the wrong one.

## `mode: monitor` is worse, not safer

The usual rollout advice ("run in monitor mode first") makes things worse here rather than mitigating:

```
monitor mode: benign=true  sqli=true   → every request is treated as a match
```

Because the engine never reports a block in monitor mode, `isBlocked()` is always false, so the inverted `return true` fires for **every** request. `testMonitorModeNeverBlocks` asserts exactly this while its name and inline comment claim the opposite:

```php
$this->assertTrue($plugin->evaluate($request), 'monitor mode should not block even on SQLi');
```

There is therefore no configuration of this plugin that is currently safe.

## Suggested fix

- [ ] `src/Plugins/Crs.php` — swap the returns: `return true` inside the `isBlocked()` branch (line 79), `return false` at the end (line 90).
- [ ] `src/Plugins/PluginInterface.php` — correct the `@return` on `evaluate()` to state that TRUE means the request **matched** and the plugin's `response:` should be applied. Consider renaming it `matches()` in 3.x to remove the ambiguity permanently.
- [ ] `tests/Unit/Plugins/CrsTest.php` — invert the `evaluate()` assertions in `testBenignRequestPasses`, `testSqliRequestBlocks`, `testXssRequestBlocks`, `testScannerUserAgentBlocks`, `testDisabledRulesAreSkipped` and the paranoia-level cases; rename `testMonitorModeNeverBlocks` and fix its assertion so monitor mode produces **no match**.
- [ ] Consider a shared contract test asserting the same polarity across every `PluginInterface` implementation, so an outlier cannot ship again.
- [ ] Release note flagging that anyone who enabled the CRS plugin should check for blocked legitimate traffic, and for attacks let through while it was active.

## Related

- #9 — where the CRS plugin was introduced.
- #74 — the 3.x merged-bucket proposal. Its worked example assumes CRS correctly blocks a SQLi payload from an allow-listed IP, which is not current behaviour.

## Environment

- `kanopi/firewall` v2.8.0 (defect identical on the `2.x` branch)
- `kanopi/crs-engine` 0.1.0
- PHP 8.4.20
- Reproduced both standalone (Composer autoloader only, no framework) and via Drupal 11.4.4

sean-e-dietrich:
## Description

Fixes #89. `Crs::evaluate()` returned the opposite boolean to every other plugin, so with `response: block` the CRS plugin **blocked benign traffic and let genuine attacks through**.

I reproduced the issue exactly as reported before changing anything:

```
-- Crs, mode: block --
  benign GET /       => true  expected false <-- WRONG   engine: action=allow score=0
  SQL injection      => false expected true  <-- WRONG   engine: action=block score=3

-- Crs, mode: monitor --
  benign GET /       => true  expected false <-- WRONG
  SQL injection      => true  expected false <-- WRONG
```

The engine was never at fault — `getLastVerdict()` was correct in every case. Only the plugin's translation of that verdict into a bool was wrong. Monitor mode was *worse* than block mode rather than safer: the engine never sets a blocking verdict there, so the inverted `return true` fired for **every** request, leaving no safe configuration of this plugin.

## Related Issue

Fixes #89

## Changes Made

**`src/Plugins/Crs.php`** — swap the two returns. Two-line behavioral change, plus comments explaining the polarity at both sites.

**`src/Plugins/PluginInterface.php`** — the root cause. The docblock said `@return TRUE if allowed to pass`, which contradicts `PluginManager::evaluate()` (treats truthy as *matched*), the other seven plugins, and the README's own custom-plugin example. CRS was the one implementation written against the docblock, so fixing only the plugin would leave the trap for the next contributor. Rewritten to state the actual contract — TRUE means the request matched and the entry's `response:` should be applied. Every plugin inherits this via `{@inheritdoc}`, so all of them are corrected.

**`tests/Unit/Plugins/CrsTest.php`** — CI stayed green because this file encoded the same misreading, asserting TRUE for benign input and FALSE for SQLi. Assertions inverted and tests renamed to say what they actually check. `testMonitorModeNeverBlocks` asserted the opposite of its own name and inline comment.

**`tests/Unit/Plugins/PluginPolarityTest.php`** (new) — a contract test giving the convention one home. Each plugin supplies a request that must match and one that must not. A coverage check fails when a plugin in `src/Plugins/` has no case, so an outlier cannot ship again. `GeoLocation` and `Asn` are exempt with a stated reason — both need MaxMind `.mmdb` files unavailable to unit tests.

## Testing

- **Verified the new test catches the original defect.** Restoring the inverted returns fails 10 tests across `PluginPolarityTest` and `CrsTest`. Without that check the contract test would be decoration.
- The issue's own reproduction script now passes every case, including both monitor-mode cases.
- 801 unit tests, 1,629 assertions — pass
- 49 integration tests, 460 assertions — pass
- PHPCS clean, PHPStan level max clean, Rector clean

One scaffolding note: my first draft of the contract test built `REQUEST_URI` with a raw space, which trips CRS protocol-enforcement rules and made a "non-matching" case match for a reason the test never intended. The helper now re-encodes the query string.

## Breaking Changes

None in the API sense — the signature is unchanged. But this **changes runtime behavior for anyone running the CRS plugin**, which is the point: CRS verdicts now apply the way the configuration says they should.

## Follow-ups not in this PR

- **Release note.** Anyone who enabled the CRS plugin should check for legitimate traffic that was blocked, and for attacks let through while it was active. Worth calling out prominently given `mode: monitor` was also affected.
- **#74** (3.x merged-bucket proposal) — its worked example assumes CRS correctly blocks a SQLi payload from an allow-listed IP, which is true as of this PR.
- The issue suggests renaming `evaluate()` to `matches()` in 3.x to remove the ambiguity permanently. Reasonable, but out of scope for a patch fix.

### DIFF
--- a/src/Plugins/Crs.php
+++ b/src/Plugins/Crs.php
@@ -68,6 +68,9 @@ class Crs extends AbstractPluginBase
         $crsVerdict = $this->getEngine()->evaluate($this->adaptRequest($request));
         $this->lastVerdict = $crsVerdict;
 
+        // TRUE means "this plugin matched", not "allow the request" — the
+        // PluginManager applies the entry's `response:` when we return TRUE.
+        // See PluginInterface::evaluate().
         if ($crsVerdict->isBlocked()) {
             $this->getLogger()->info('CRS blocked request', $this->getContext($request, [
                 'rule_id'      => $crsVerdict->blockingRuleId,
@@ -76,9 +79,11 @@ class Crs extends AbstractPluginBase
                 'matched_rule' => $crsVerdict->matchedRules[0]['msg'] ?? '',
                 'matched_data' => $crsVerdict->matchedRules[0]['matched_data'] ?? '',
             ]));
-            return false;
+            return true;
         }
 
+        // Rules fired but the anomaly score stayed under threshold, or the
+        // plugin is in monitor mode. Report no match so the request proceeds.
         if ($crsVerdict->matchedRules !== []) {
             $this->getLogger()->debug('CRS matched but did not block', $this->getContext($request, [
                 'action'          => $crsVerdict->action,
@@ -87,7 +92,7 @@ class Crs extends AbstractPluginBase
             ]));
         }
 
-        return true;
+        return false;
     }
 
     /**
--- a/src/Plugins/PluginInterface.php
+++ b/src/Plugins/PluginInterface.php
@@ -29,13 +29,26 @@ interface PluginInterface
     public function getDescription(): string;
 
     /**
-     * Evaluate the current request to see if it passes the current plugin.
+     * Evaluate the current request against this plugin's rules.
+     *
+     * This reports whether the plugin **matched**, not whether the request
+     * should be allowed. What a match does is decided by the entry's
+     * `response:` setting, which PluginManager applies:
+     *
+     * - `response: allow`     — a match permits the request immediately.
+     * - `response: challenge` — a match serves the interstitial.
+     * - `response: block`     — a match rejects the request.
+     *
+     * So a plugin that detects something bad returns TRUE. Returning FALSE
+     * means "my rules did not apply to this request", which lets evaluation
+     * continue to the next plugin.
      *
      * @param Request $request
      *   Request to evaluate.
      *
      * @return bool
-     *   Return TRUE if allowed to pass.
+     *   TRUE if the request matched this plugin's rules and the configured
+     *   `response:` should be applied; FALSE if it did not match.
      */
     public function evaluate(Request $request): bool;
 
--- a/tests/Unit/Plugins/CrsTest.php
+++ b/tests/Unit/Plugins/CrsTest.php
@@ -17,6 +17,11 @@ use Symfony\Component\HttpFoundation\Request;
  * mapping, configuration plumbing, status/expiration accessors, and the
  * bool return contract of evaluate().
  *
+ * That contract is "did this plugin match", not "should the request pass" —
+ * see PluginInterface::evaluate(). A payload CRS blocks makes evaluate()
+ * return TRUE. PluginPolarityTest enforces the same convention across every
+ * shipped plugin.
+ *
  * Tests skip themselves if the engine's bundled rules haven't been generated
  * (vendor install pulls the rules cache in, but a fresh dev clone without
  * `composer install` won't have them).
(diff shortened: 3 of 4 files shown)


kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 96.11
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/e7830e4b-92ec-4813-b702-477a1f99bb6b/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/b1746b1e-1785-4f19-a3e9-91916ef86c58/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/2dfd7b4a-876d-4b0e-bb2e-6b0c54e4613f/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 96.11
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/a6e723a4-4649-4a1a-9dfb-efd6ba4a07a4/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/94073b2f-2c7c-4be7-971d-2dd4536d90db/artifacts/0/reports/index.html)

PR #90, opened 2026-07-26
Add MkDocs Material documentation site

sean-e-dietrich:
## Description

Stands up a documentation site at **kanopi.github.io/firewall**, built from Markdown in `docs/` and published to GitHub Pages by CircleCI.

The configuration reference had grown to a 3,063-line README — hard to navigate, harder to contribute to. This splits it (plus the preset, example, and contributing docs) into 45 focused pages.

**Nothing was dropped.** 5,796 lines of docs out of 5,332 lines of source.

## Why MkDocs Material

Contributors only ever touch `.md` files and a `nav:` list — no Node or Ruby toolchain lands in a PHP repo. Every page carries an edit pencil that opens the file in GitHub's web editor and walks a contributor through fork → commit → PR without leaving the browser.

## Changes Made

**Content** — 28 pages split from `README.md`, 16 from `presets/`, `example/`, and `CONTRIBUTING.md`. All ~50 cross-references rewritten for the new layout. Source files slimmed to pointers so nothing is documented twice:

| File | Before | After |
|---|---|---|
| `README.md` | 3,063 | 165 |
| `presets/README.md` | 890 | 52 |
| `example/README.md` | 505 | 43 |
| `CONTRIBUTING.md` | 402 | 70 |
| `presets/RATE-LIMITING-REFERENCE.md` | 336 | 13 |
| `example/demo/README.md` | 136 | 27 |

**Screenshots** — four captured from the demo app (math + ALTCHA interstitials, block response, allowed page), shown as content tabs on the challenge and demo pages.

**Tooling** — `mkdocs.yml` enables synced per-framework content tabs, code annotations, Mermaid, and `pymdownx.snippets`, so examples can embed real shipped files (`presets/*.yml`) rather than copies that drift. `docs/requirements.txt` pins the toolchain; `pymdown-extensions` is held below 11 deliberately — it installs over mkdocs-material's `~=10.2` constraint without complaint, then fails at render time.

**CI** — `docs-build` runs on every branch and PR and stores rendered HTML as an artifact with a preview link commented on the PR. `docs-deploy` publishes the reviewed artifact to `gh-pages`, gated on a **stable release tag** (`^v\d+\.\d+\.\d+$`) rather than on a merge to `2.x` — see "How publishing works" below. The push credential is set on the remote rather than the command line so a git failure cannot echo the token into the build log.

## Testing

- `mkdocs build --strict` passes — 45 pages, 0 warnings
[...]

PR #92, opened 2026-07-27
Pin phpstan below 2.2.6 to unbreak Rector in CI

sean-e-dietrich:
## Description

Every `rector` job on every branch is currently failing. It is **not** a lint finding and **not** caused by any open PR — Rector dies with a fatal error before it reads a single file:

```
Rector\Exception\Reflection\MissingPrivatePropertyException:
Property "$container" was not found in "PHPStan\Parser\RichParser"
  in vendor/rector/rector/src/Util/Reflection/PrivatesAccessor.php:82
```

Rector's `PHPStanContainerMemento::removeRichVisitors()` reaches into PHPStan's `RichParser` for a private `$container` property via reflection. **phpstan 2.2.6, published 2026-07-26, removed it.** Rector's latest release (2.5.7, 2026-07-13) predates the change, so there is nothing to upgrade to yet.

## Evidence

| phpstan | rector | result |
|---|---|---|
| 2.2.5 | 2.5.7 | ✅ 53/53 files, `[OK] Rector is done!` |
| 2.2.6 | 2.5.7 | ❌ fatal, identical to CI |

Reproduced locally by upgrading **only** phpstan and changing nothing else.

Confirmed unrelated to any branch's code: **PR #90 touches zero PHP files and fails rector identically** on all five PHP versions.

## Changes Made

One line in `composer.json`:

```diff
-"phpstan/phpstan": "^2.1",
+"phpstan/phpstan": "^2.1 <2.2.6",
```

Drop the upper bound once Rector ships a compatible release.

## Why pin rather than commit `composer.lock`

A lock file would pin one dependency resolution for all five PHP versions, which defeats the point of the matrix — the whole reason `composer.lock` is gitignored here is so each PHP version resolves the versions appropriate to it. Pinning the one broken package keeps that intact.

The tradeoff to accept knowingly: because CI resolves fresh on every run, a third-party patch release can turn the build red with no commit involved. That is what happened here. The mitigation is narrow upper bounds on dev tools that reach into another tool's internals — Rector↔PHPStan is the fragile coupling, since Rector drives PHPStan's container by reflection rather than through a public API. It is the one pairing in this repo likely to break this way again.

## Testing

With the pin applied:

- `rector process src/ --dry-run --clear-cache` — 53/53 files, clean
- `phpstan analyse src --level max` — no errors
- `phpcs` — clean
- 788 unit tests pass
- `composer why-not phpstan/phpstan 2.2.6` confirms the constraint excludes the broken release
- A fresh resolve (what CI does) selects 2.2.5

Verified the pin is safe across the whole matrix, not just locally:

- phpstan 2.2.5 requires `php: ^7.4|^8.0`, so it installs identically on 8.1 through 8.5 — no PHP-version divergence for this package
- rector 2.5.7 requires `phpstan/phpstan: ^2.2.2`, which 2.2.5 satisfies — no dependency conflict introduced
- `composer update --dry-run` reports zero conflicts

## Breaking Changes

None. Dev dependency constraint only; nothing in `require` and no runtime code touched.

## Merge order

This should go in first — it unblocks #90 and #91, which both need a rebase (or a merge from `2.x`) afterwards to go green.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^1.0",
         "dg/bypass-finals": "^1.9",
         "phpcompatibility/php-compatibility": "^9.3",
-        "phpstan/phpstan": "^2.1",
+        "phpstan/phpstan": "^2.1 <2.2.6",
         "phpunit/phpunit": "^10.5 || ^11.5",
         "rector/rector": "^2.0",
         "squizlabs/php_codesniffer": "^3.13",


kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/e752b3b4-b6d5-416e-9d41-6b37483d278f/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/190d541b-4e1a-44c7-9f0e-3ef68870731d/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 96.11
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/54586a28-850c-44d9-9599-afa033685d63/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 96.11
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/c7ca88e6-9891-4f33-8c94-3a2161761e07/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/2adb623c-18b2-4daf-8869-4b74b5782de0/artifacts/0/reports/index.html)

PR #95, opened 2026-07-27
Name the CRS anomaly thresholds honestly (Fixes #93)

### LINKED ISSUE #93: anomaly_thresholds is misnamed: two of four keys are ignored, and the threshold is bypassed by most rules
## Summary

`anomaly_thresholds` does not behave the way its name and documentation imply, in two separate ways:

1. **Two of its four keys are read nowhere.** `warning` and `notice` are accepted, documented, and silently ignored.
2. **`critical` is not a per-severity threshold.** It is the single global inbound anomaly score threshold. `error` is the outbound equivalent.

Separately, and more consequentially for anyone tuning the rule set: **the threshold is not consulted for the majority of rules**, because a rule carrying a `block`/`deny`/`drop` action rejects the request the moment it matches. So raising the threshold is not the false-positive lever the documentation suggests it is.

This is a documentation and naming problem rather than a security one — nothing here causes wrong allow/block decisions. But it does send operators down a dead end when they are trying to tune away a false positive, which is the moment they are least able to afford it.

## Where the keys go

`kanopi/crs-engine` `src/Runtime/CrsTxDefaults.php`:

```php
// Per-severity score contributions — hardcoded, not configurable.
'tx.critical_anomaly_score' => '5',
'tx.error_anomaly_score'    => '4',
'tx.warning_anomaly_score'  => '3',
'tx.notice_anomaly_score'   => '2',

// Thresholds.
'tx.inbound_anomaly_score_threshold'  => (string) ($crsConfig->anomalyThresholds['critical'] ?? 5),
'tx.outbound_anomaly_score_threshold' => (string) ($crsConfig->anomalyThresholds['error'] ?? 4),
```

So:

| Config key | Actually controls |
|---|---|
| `anomaly_thresholds.critical` | `tx.inbound_anomaly_score_threshold` — the global inbound threshold |
| `anomaly_thresholds.error` | `tx.outbound_anomaly_score_threshold` — unused until response-side evaluation lands (#69) |
| `anomaly_thresholds.warning` | nothing |
| `anomaly_thresholds.notice` | nothing |

`grep -rn anomalyThresholds src/` in `crs-engine` returns only those two lines plus one read in `RuleEvaluator`. The defaults `['critical' => 5, 'error' => 4, 'warning' => 3, 'notice' => 2]` happen to match the CRS *per-severity score contributions*, which is almost certainly where the confusion originated — those numbers belong to `tx.*_anomaly_score`, not to any threshold.

## Why the threshold rarely applies

`crs-engine` `src/Runtime/RuleEvaluator.php`:

```php
if ((in_array($compiledRule->action, ['deny', 'block', 'drop'], true)) && $this->crsConfig->mode === CrsConfig::MODE_BLOCK) {
    $blockingId = $compiledRule->id;
    break;
}
```

and:

```php
private function decideAction(?int $blockingId, int $totalScore): string
{
    if ($blockingId !== null && $this->crsConfig->mode === CrsConfig::MODE_BLOCK) {
        return CrsVerdict::ACTION_BLOCK;      // threshold never consulted
    }

    $threshold = $this->crsConfig->anomalyThresholds['critical'] ?? PHP_INT_MAX;
    // ...
}
```

Counting actions across the shipped compiled rule set:

```
block    338
pass     241
deny       4
---------------
block/deny/drop: 342 of 583 (59%)
```

So for 59% of rules the first match short-circuits and the threshold is bypassed. That includes the rules operators actually meet: SQL injection and XSS both reject outright.

Demonstrated — raising the threshold by two orders of magnitude changes nothing:

```php
foreach ([5, 50, 500] as $t) {
    $p = new Crs([], ['paranoia' => 1, 'mode' => 'block',
        'anomaly_thresholds' => ['critical' => $t, 'error' => 4, 'warning' => 3, 'notice' => 2]]);
    var_dump($p->evaluate(Request::create('/', 'GET', ['id' => "1' UNION SELECT 1,2,3--"])));
}
// true, true, true — blocked by rule 942190 every time, score 5
```

The same holds for an XSS payload, blocked by 941120.

This is a defensible design — it mirrors ModSecurity's `deny` actions — but it differs from stock CRS, where blocking is normally routed through the anomaly-evaluation rules (949110 / 959100) and the threshold genuinely governs everything. Worth being explicit about, because operators arriving from a ModSecurity background will expect the CRS behaviour.

## Suggested changes

Any one of these would help; the first two are the cheapest.

- [ ] **Document it.** In the CRS plugin section of the README, state that `anomaly_thresholds.critical` is the inbound threshold, that `error` is outbound and currently unused, that `warning` and `notice` are ignored, and that the threshold only applies to requests which trip no outright-rejecting rule. Point operators at `disabled_rules` as the false-positive lever.
- [ ] **Rename the key** to something honest, e.g. `anomaly_score_thresholds: {inbound: 5, outbound: 4}`, keeping the old shape accepted with a deprecation. Two keys that do something beats four keys where two are inert.
- [ ] **Reject or warn on the inert keys.** Passing `warning` or `notice` today looks like it worked. `CrsConfig` could emit a `LoggingFactory` warning, or throw `ConfigurationException` on unknown keys.
- [ ] **Optionally, honour the score contributions too.** `tx.critical_anomaly_score` and friends are hardcoded; if they were configurable, the four-key shape would become meaningful rather than misleading.

## Impact on consumers

The Drupal module built on this library (`basic_firewall`) previously rendered all four keys as editable per-severity thresholds with the description "raising a threshold makes that severity less likely to trigger a block" — wrong on both counts, and written straight from the key names and README. It now exposes only `critical` and `error`, labelled inbound and outbound, and states that raising the threshold will not make the rule set passive. That correction was only possible after reading the engine source, which is the sort of thing the README should make unnecessary.

## Environment

- `kanopi/firewall` v2.8.1
- `kanopi/crs-engine` 0.1.0
- PHP 8.4.20

sean-e-dietrich:
> ### Updated for `crs-engine` 1.0.0
>
> Since this PR was opened, 1.0.0 shipped and fixed the upstream defects described below — including the `inbound`/`outbound` rename this PR proposed downstream (crs-engine#14) and the dead anomaly-evaluation path (crs-engine#13, crs-engine#1). **The threshold now works.** The analysis below is preserved as the record of why #93 was filed and what 0.1.0 actually did; where it says the threshold is inert, that is history. See [What changed for 1.0.0](#what-changed-for-100) at the bottom for the current state of the branch.

Fixes #93.

## The issue is real

Verified against `kanopi/crs-engine` 0.1.0 as vendored:

- `anomaly_thresholds` accepts four keys and reads two. `critical` is the single **inbound** threshold, `error` the **outbound** one. `warning` and `notice` are read nowhere — `grep -rn anomalyThresholds src/` in the engine returns only the two lines in `CrsTxDefaults` and one in `RuleEvaluator`.
- The threshold is bypassed whenever a rule carrying a `block` / `deny` / `drop` action matches, because `RuleEvaluator` breaks on first such match and `decideAction()` returns `block` without consulting the score.

**It is worse than the issue reports.** The issue puts it at 59% of rules. Scanning the bundled compiled ruleset for rules that increment `tx.inbound_anomaly_score_pl*` — the only thing `totalScore()` sums:

```
block: 208
pass:    0
```

Every score-contributing rule carries a `block` action. So in block mode the first one short-circuits and the threshold is never compared; in monitor mode nothing blocks and both sides of the comparison return `log`. **The inbound threshold has no observable effect on any verdict with the shipped rule set.** Confirmed end to end:

```
block    t=5       action=block  score=5   rule=942190
block    t=50      action=block  score=5   rule=942190
block    t=500     action=block  score=5   rule=942190
monitor  t=1       action=log    score=10
monitor  t=100000  action=log    score=10
```

## What this PR changes

The engine lives in a separate package, so the rename lands at the plugin's config surface — which is what firewall users actually write.

**`src/Plugins/Crs.php`**
- Accept `inbound` / `outbound` and map them onto the engine's `critical` / `error`. Severity spellings stay accepted so existing config keeps working; the preferred spelling wins when both are present, so config migrated one key at a time behaves the same in either write order.
- Any other key — including the `warning` / `notice` our own README told people to write — logs a warning naming what it ignored and pointing at `disabled_rules`, instead of looking like it worked.
- A non-numeric threshold, or a non-mapping `anomaly_thresholds`, falls back to the default and says so rather than silently casting to `0` (which would have made the engine maximally strict).

**`README.md`** — new *Anomaly scores and thresholds* section: which two keys are read, that per-severity weights are fixed by CRS and not configurable, that the threshold is inert against the shipped rule set, and the four settings that actually do move a false positive (`disabled_rules`, `disabled_categories`, `paranoia`, `mode: monitor`).

**`example/config.notes.yml`** — same correction.

**`tests/Unit/Plugins/CrsTest.php`** — 10 new tests. Key mapping, precedence, defaults, the inert-key warning, non-numeric and non-mapping fallbacks, plus two that pin the inertness claim in block and monitor mode. If the engine ever routes blocking through the anomaly-evaluation rules (949110 / 959100) the way stock CRS does, those two fail — which is the signal to revisit the README section rather than weaken the tests.

## Root cause is upstream, and it is not a design choice

The issue frames the short-circuit as deliberate — "it mirrors ModSecurity's `deny` actions" rather than stock CRS. That is not what is happening. CRS's anomaly-evaluation rules **are** in the bundled set:

```
949111 action=deny op=ge arg=%{tx.inbound_anomaly_score_threshold}  target=BLOCKING_INBOUND_ANOMALY_SCORE
  chained 949110 ... same comparison, phase 2
959101 action=deny op=ge arg=%{tx.outbound_anomaly_score_threshold} target=BLOCKING_OUTBOUND_ANOMALY_SCORE
  chained 959100 ...
```

They can never fire. The aggregator rules that feed `TX.BLOCKING_INBOUND_ANOMALY_SCORE` (949052–949063) have their `setvar` values compiled to a literal `0`, and the flattening happens at parse time — `REQUEST-949-BLOCKING-EVALUATION.json` already carries `0`, before `compiled.php` is built:

```json
{ "id": 949052, "setvars": [ { "name": "tx.blocking_inbound_anomaly_score", "op": "+", "value": "0" } ] }
```

`crs-engine`'s parser inlines `%{tx.*}` macros at build time. That is correct for fixed seeds — `%{tx.critical_anomaly_score}` becomes `"5"`, as in 930120 — but wrong for the runtime accumulators `%{tx.inbound_anomaly_score_plN}`, which have no seed in `CrsTxDefaults` and resolve to zero. Across the shipped ruleset, **no compiled setvar retains a macro at all** and 32 increment by a literal `0`.

With its input pinned at `0`, the entire anomaly-evaluation path is dead and blocking falls back to per-rule `block` actions — exactly the symptom #93 reports. Note the threshold macro *is* preserved in the operator argument and resolved at runtime, so the moment the parser keeps those setvar macros, `anomaly_thresholds.inbound` goes live with no config change on this side. That is the argument for keeping the key rather than deprecating it.

## Not done here

The parser defect above is the real fix and belongs in `kanopi/crs-engine`. So do two of the issue's other suggestions: throwing `ConfigurationException` on unknown keys in `CrsConfig`, and making the per-severity score contributions (`tx.critical_anomaly_score` and friends) configurable. Nothing in this repo can address them — `crs-engine` currently has no issues filed; happy to open one with the reproduction above.

## Drive-by

Both stale since #91 inverted `Crs::evaluate()`, both in sections this PR was already editing:

- README's `getLastVerdict()` example said "evaluate() returns FALSE when CRS would block the request" and assigned it to `$allowed`.
- `config.notes.yml` described `mode: block` as "returns false from evaluate()".

## Verification

811 unit tests and 49 integration tests pass. PHPCS, PHPStan at level max, and Rector are clean.

---

## What changed for 1.0.0

Three things, none cosmetic. Verified against 1.0.0 in an isolated sandbox and a throwaway worktree before touching the working tree.

### It would not install

`"kanopi/crs-engine": "^0.1.0"` excludes 1.0.0 — `composer prohibits` confirms. Bumped to `^1.0`. Not worth supporting 0.1.0 alongside; the engine's own release notes call its detection pipeline "substantially inert".

### The shim emitted deprecated keys on every request

1.0.0 keys thresholds `inbound`/`outbound` and emits `E_USER_DEPRECATED` for the old severity spellings. This plugin's `anomalyThresholds()` normalised *to* `critical`/`error`, so it took the alias path — two deprecations per plugin instance, for **every** user, including those with no `anomaly_thresholds` config at all, since we always passed both keys explicitly. It now emits the canonical names:

```
config #0 []                          deprecations=0
config #1 {"inbound":7}               deprecations=0
config #2 {"critical":7}              deprecations=0
config #3 {"critical":7,"warning":3}  deprecations=0
```

Legacy config keeps working, and doing the translation here keeps the engine's deprecation out of the host application's logs.

### The threshold is now the lever, and the docs said the opposite

`testRaisingTheThresholdDoesNotDisarmBlockingRules` was written in this PR to fail when this got fixed. It did — the single failure across 811 unit + 49 integration tests run against 1.0.0:

```
inbound=5    matched=true   action=block  score=10  rule=949110
inbound=50   matched=false  action=log    score=10  rule=-
inbound=500  matched=false  action=log    score=10  rule=-
```

Replaced with its inverse: the same SQLi payload is rejected at threshold 5 and allowed at 500, same score either way. README, the class docblock and `config.notes.yml` rewritten accordingly.

### The sharpest edge, which the rename does not cover

Blocking now routes through CRS rule **949110**, so `rule_id` is 949110 on every anomaly-score block rather than the rule that caught the payload. The old README told operators to take `rule_id` and put it in `disabled_rules` — following that today disables anomaly blocking wholesale.

Added `contributing_rules` to the block and debug log lines (detections only, `blocking_evaluation` dropped) and repointed all tuning guidance at it. Also logging the new `CrsVerdict::$operatorErrors` / `$truncations`, at warning level even when nothing matched — a clean verdict over a partly-uninspected request is exactly where a coverage gap gets believed.

### Smaller items

- Category list gained `blocking_evaluation`, `response_leak`, `response_leak_ruby`. Documented, with `blocking_evaluation` flagged off-limits for `disabled_categories` for the same reason as 949110.
- The plugin stays more forgiving than the engine, now deliberately: 1.0.0 throws `ConfigurationException` on an unknown threshold key, this plugin drops it with a warning. A YAML typo should not take a site down. Stated in the docblock so it reads as a choice.
- Upgrade note added to the README. Traffic that passed under 0.1.0 will now be rejected, and CRS at PL1 flags some ordinary editorial content — a support ticket containing `cat /etc/hosts | grep` scores 15 and is rejected. Default stays `mode: block` with monitor-first rollout documented, rather than silently weakening an existing deployment's protection.

### Verification

812 unit + 49 integration tests pass. PHPCS, PHPStan at level max (`composer check:security`), and Rector clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1863,15 +1863,48 @@ plugins:
       #   response_leak_java, response_leak_php, response_leak_iis,
       #   web_shell, correlation
       disabled_categories: []
-      # Override anomaly-score thresholds
+      # Anomaly-score thresholds. There are exactly two — see
+      # "Anomaly scores and thresholds" below before reaching for these.
       anomaly_thresholds:
-        critical: 5
-        error: 4
-        warning: 3
-        notice: 2
+        # Request-side score at or above which the request is rejected.
+        inbound: 5
+        # Response-side equivalent. Inert until response evaluation lands (#69).
+        outbound: 4
       # Custom rule cache location — defaults to vendor/kanopi/crs-engine/rules
       # rules_path: /etc/firewall/crs-rules
 ```
+
+#### Anomaly scores and thresholds
+
+A matching rule adds points to the request's anomaly score, weighted by the rule's CRS severity — critical contributes 5, error 4, warning 3, notice 2. **Those per-severity weights are fixed by CRS and are not configurable here.**
+
+What *is* configurable is the two thresholds the score is compared against:
+
+| Key | Default | Controls |
+|---|---|---|
+| `inbound` | `5` | Request-side threshold — the score at or above which a request is rejected. Read the caveat below before tuning it. |
+| `outbound` | `4` | Response-side equivalent. Inert until response evaluation lands (issue #69). |
+
+The severity-named spellings `critical` (= `inbound`) and `error` (= `outbound`) are still accepted so existing config keeps working. Any other key — including `warning` and `notice`, which earlier versions of this document showed — is read nowhere; the plugin now logs a warning naming the keys it ignored instead of accepting them silently. If both spellings of a threshold are present, `inbound` / `outbound` win.
+
+**The threshold is not the false-positive lever it looks like.** A rule whose CRS action is `block`, `deny`, or `drop` rejects the request the moment it matches, without consulting the score at all — and in the rule set shipped with `crs-engine` today, *every* rule that contributes to the anomaly score carries such an action. The score therefore never reaches the threshold comparison in block mode, and in monitor mode nothing blocks either way. Raising `inbound` from 5 to 500 changes nothing:
+
+```php
+foreach ([5, 50, 500] as $threshold) {
+    $crs = new Crs([], ['mode' => 'block', 'anomaly_thresholds' => ['inbound' => $threshold]]);
+    var_dump($crs->evaluate(Request::create('/', 'GET', ['id' => "1' UNION SELECT 1,2,3--"])));
+}
+// true, true, true — blocked by rule 942190 every time, score 5.
+```
+
+This mirrors ModSecurity's `deny` actions rather than stock CRS, where blocking is routed through the anomaly-evaluation rules (949110 / 959100) and the threshold governs everything. The setting is honoured by the engine and matters for rule sets whose score-contributing rules use non-blocking actions (`rules_path`), but for the bundled set treat it as inert.
+
+So, to tune a false positive:
+
+- **`disabled_rules`** — the precise lever. The blocked request's log line carries `rule_id`; add it here.
+- **`disabled_categories`** — blunter, when a whole class of rules is wrong for your app.
+- **`paranoia`** — lower it to drop the aggressive rule tiers wholesale.
+- **`mode: monitor`** — the only setting that reliably stops CRS rejecting anything while you tune.
 
 #### Coverage
 
@@ -1907,8 +1940,9 @@ $crs = new Crs(metadata: [], config: [
     'mode' => 'monitor',   // Score and report without blocking.
 ]);
 
-// evaluate() returns FALSE when CRS would block the request.
-$allowed = $crs->evaluate($request);
+// evaluate() returns TRUE when the plugin matched — i.e. when CRS would
+// block the request. See PluginInterface::evaluate().
+$matched = $crs->evaluate($request);
 
 $verdict = $crs->getLastVerdict();
 if ($verdict !== null) {
--- a/README.md
+++ b/README.md
@@ -1897,7 +1897,9 @@ foreach ([5, 50, 500] as $threshold) {
 // true, true, true — blocked by rule 942190 every time, score 5.
 ```
 
-This mirrors ModSecurity's `deny` actions rather than stock CRS, where blocking is routed through the anomaly-evaluation rules (949110 / 959100) and the threshold governs everything. The setting is honoured by the engine and matters for rule sets whose score-contributing rules use non-blocking actions (`rules_path`), but for the bundled set treat it as inert.
+This is not a deliberate departure from stock CRS. The anomaly-evaluation rules that normally do the blocking are present in the bundled set — 949111 chained with 949110 inbound, 959101 with 959100 outbound — and they do compare `TX.BLOCKING_INBOUND_ANOMALY_SCORE` against `%{tx.inbound_anomaly_score_threshold}` with a `deny` action. They can never fire, because the aggregator rules that feed that variable (949052–949063) have their `setvar` values compiled to a literal `0`: `crs-engine`'s parser inlines `%{tx.*}` macros at build time, which is right for fixed seeds like `%{tx.critical_anomaly_score}` but flattens the runtime accumulators `%{tx.inbound_anomaly_score_plN}` — which have no seed value — to zero. With its input pinned at 0 the whole anomaly-evaluation path is dead, and blocking falls back entirely to per-rule `block` actions.
+
+That is an upstream defect in `kanopi/crs-engine`, not something this plugin can fix. Note the threshold macro *is* preserved in the operator argument and resolved at runtime, so once the parser keeps those setvar macros, `inbound` becomes live and meaningful with no config change here.
 
 So, to tune a false positive:
 
--- a/README.md
+++ b/README.md
@@ -1836,6 +1836,12 @@ Evaluates each request against the [OWASP Core Rule Set](https://coreruleset.org
 - **In-process rule cache**: ~3-4 ms per request once warm (FPM worker steady state). Zero extension dependencies — no APCu / OPcache preload required.
 - **Auto-refreshed rules**: The `crs-engine` package CI fetches new CRS releases weekly and opens a reviewable PR; `composer update` pulls the latest curated bump.
 
+> ### Upgrading from a release built on `crs-engine` 0.1.0
+>
+> This plugin now requires `crs-engine` ^1.0, which is the first version where the engine actually enforces. In 0.1.0 the detection pipeline was substantially inert — the anomaly-evaluation rules could never fire, `TX:` targets never resolved, and `skipAfter` silently discarded the rest of the ruleset. **Traffic that used to pass will now be rejected**, because previously very little was.
+>
+> Roll it out with `mode: monitor` first. Let it run over real traffic, group the `contributing_rules` on the debug log lines by rule ID, and build your `disabled_rules` list from what you actually see before you let it block. CRS at paranoia 1 does flag some ordinary editorial content — a support ticket containing `cat /etc/hosts | grep` scores 15 and would be rejected — so on a CMS this step is not optional.
+
 #### Configuration Example
 
 ```yaml
@@ -1859,9 +1865,11 @@ plugins:
       # Or by category. Available categories:
       #   sqli, xss, lfi, rfi, rce, php, java, session_fixation,
       #   protocol_attack, protocol_enforcement, method_enforcement,
-      #   scanner, multipart, generic, response_leak_sql,
+      #   scanner, multipart, generic, response_leak, response_leak_sql,
       #   response_leak_java, response_leak_php, response_leak_iis,
-      #   web_shell, correlation
+      #   response_leak_ruby, web_shell, correlation, blocking_evaluation
+      # Do not disable blocking_evaluation — it is how anomaly scoring
+      # rejects anything at all.
       disabled_categories: []
       # Anomaly-score thresholds. There are exactly two — see
       # "Anomaly scores and thresholds" below before reaching for these.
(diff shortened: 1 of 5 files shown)


kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 96.17
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/ffeb81d3-9e30-4672-b02e-aaa473c15080/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 96.17
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/6c2d7773-66ba-411f-8e8e-f222640e6d6a/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 96.17
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/3ef8d0bf-e1f1-4e79-ace6-e45a831171cc/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/292d4aa8-1eb2-4f06-9588-3515bf92a4e3/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 96.14
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/91147474-e8d0-481d-9944-290373903897/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 96.04
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/52719628-c80d-4679-9714-a1802eec4057/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 96.01
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/9837dde5-3e5e-467e-b28e-1502bf92e08b/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 96.01
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/bc7a65dc-e1c3-4182-8f22-8b9fcf6b3e9c/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 96.04
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/93e83f51-bd81-4daf-a734-42b1254cdc1d/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 96.04
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/c8bb2e7c-b62f-41ed-af1c-6c9099b78f7b/artifacts/0/reports/index.html)

PR #106, opened 2026-07-30
Add bin/firewall-check to test a request against a config (Fixes #105)

### LINKED ISSUE #105: Add a CLI to test whether a given request would be blocked
## Summary

There is no way to ask "would this request be blocked, and by what?" without either sending real traffic at a running site or writing a throwaway PHP script.

Everything needed already exists — `Firewall::evaluate()` takes a `Request`, and `PluginManager::evaluate()` already records every plugin's result and timing — but assembling it correctly has sharp edges that are easy to get wrong in a way that *silently reports the wrong answer*.

## What exists today, and why it is not enough

| Approach | Limitation |
|---|---|
| `docs/presets/testing.md` curl recipes | Needs a running site; cannot test a config before deploying it |
| `mode: log` | Dry-run over **real** traffic; cannot probe a hypothetical request |
| `tests/Performance/bin/validate-scenarios.php` | Closest thing, but hardcoded to the perf scenarios and its own probe list |
| `example/test-vulnerability-score.php` | Single plugin, instantiated directly, not the real evaluation path |
| Hand-rolled script | Works, but see the traps below |

## The traps

These are the reason this belongs in the repo rather than in everyone's notes:

1. **`Firewall::evaluate()` silently bypasses in CLI.** It returns `true` immediately when `PHP_SAPI === 'cli'` unless the mode is `exception`. A checker that forgets this reports **"allowed" for every request**, including ones that obviously block. A test harness that always says "nothing is blocked" is worse than no harness.

2. **Overrides use PropertyAccess bracket paths, not nested arrays.** `['global' => ['mode' => 'exception']]` is silently ignored; it has to be `['[global][mode]' => 'exception']`. This fails quietly and produces trap 1.

3. **A block has side effects.** The block path writes to storage, records an offense, and applies `blocking_escalation`. Pointed at production config, a "check" would **actually ban the address it was asked about**.

4. **The exception does not say which plugin matched.** `FirewallBlockedException` carries only a message and status code. Attribution is available, but only from the log line `Request blocked by plugin` (`Firewall.php:1174`).

## Proposal

`bin/firewall-check`, taking the request as flags and printing the verdict, the responsible plugin, and why:

```bash
bin/firewall-check --config=firewall.yml --ip=203.0.113.5 --url=/wp-admin/
bin/firewall-check --config=firewall.yml --url='/search?q=1%27+UNION+SELECT' --explain
bin/firewall-check --config=firewall.yml --ip=1.1.1.1 --header='User-Agent: sqlmap/1.8' --json
```

- `--config` repeatable, merged in order, as `Firewall::create()` accepts.
- `--ip`, `--url`, `--method`, `--header` (repeatable), `--body`.
- `--explain` lists **every** plugin that evaluated with its result and timing, rather than stopping at the first match. This needs no extra evaluation: `PluginManager::evaluate()` already logs `evaluated_plugins` with `plugin`, `result` and `time_ms` in both the matched and unmatched branches.
- `--json` for scripting and CI.
- Storage is replaced with a throwaway by default so a check can never ban anyone; an opt-in flag restores the configured backend for those who want the durable blocklist consulted.
- Meaningful exit codes so it composes in scripts and CI.

## Why it is worth having

- **Test a config before deploying it.** Today the first real feedback on a ruleset is production traffic.
- **Answer a support ticket.** "Customer says they were blocked" becomes reproducible instead of guesswork.
- **Tune false positives.** The CRS docs are emphatic that paranoia 1 flags ordinary editorial content and that monitor-first rollout is not optional; `--explain` makes that tuning loop fast and local.
- **Pairs with #26.** `find()` / `deleteMatching()` answers "who is blocked?"; this answers "would this be blocked, and by what?". Together that is the management foundation #26 refers to.

sean-e-dietrich:
Fixes #105.

Answers **"would this request be blocked, and by what?"** from a terminal, against a config file, with no running site.

```console
$ vendor/bin/firewall-check --config=firewall.yml --ip=203.0.113.5 --url=/wp-admin/
BLOCKED  GET /wp-admin/
  client            203.0.113.5
  blocked by        IP Address
  status            400
  reason            blocked
  storage           throwaway (in-memory)
```

Everything needed already existed — `Firewall::evaluate()` takes a `Request`, and `PluginManager::evaluate()` already records each plugin's result and timing. What did not exist was a *correct* way to call it, because three things fail **silently**.

## The three traps this exists to close

**1. `evaluate()` bypasses in CLI.** It returns `true` immediately when `PHP_SAPI === 'cli'` unless the mode is `exception`. A checker that misses this reports "allowed" for every request, including ones plainly inside a blocked CIDR — worse than no checker, because it looks like it works. Mode is forced, and a test asserts a known-blocked address is not reported as allowed.

**2. Overrides are bracket paths, not nested arrays.** `['global' => ['mode' => 'exception']]` is silently ignored; it has to be `['[global][mode]' => 'exception']`. Not hypothetical — it's exactly what happened on my first throwaway script, which cheerfully reported ALLOW for four requests that all block.

**3. A block is not read-only.** It writes to storage, records an offense, and applies `blocking_escalation`. Pointed at a production config, a naive "check" **bans the address it was asked about**.

## Safety

Storage is swapped for a throwaway by default, so a check cannot ban anyone. A test asserts the configured backend's file is never created even on a blocked verdict.

`--live-storage` opts back in when you genuinely need the durable blocklist consulted, warns on stderr, and keeps stdout parseable so `--json` still works.

## `--explain`

Shows what ran, and — from the parsed config — what an earlier match short-circuited past. The second list is usually the real answer to *"why wasn't this caught?"*:

```console
$ vendor/bin/firewall-check --config=firewall.yml --ip=203.0.113.5 --url=/wp-admin/ --explain
...
Plugins evaluated, in order:
  MATCH   IP Address                     0.06 ms

Configured but not reached:
  block     Kanopi\Firewall\Plugins\Url              weight -10
  block     Kanopi\Firewall\Plugins\UserAgent        weight 0
```

Attribution is read from the firewall's own decision log rather than by re-running plugins — a second pass would repeat every side effect, including any outbound reputation lookup.

## Exit codes

`0` allowed · `1` blocked · `2` challenged · `64` usage · `70` internal. The verdict *is* the exit status, so it composes:

```bash
if vendor/bin/firewall-check --config=firewall.yml --url=/wp-admin/ >/dev/null; then
  echo "Not blocked — the WordPress preset isn't doing its job."; exit 1
fi
```

Registered under composer `bin`, so consumers get `vendor/bin/firewall-check`.

---

# How to test

## 1. Automated

```bash
vendor/bin/phpunit -c phpunit.xml --testsuite unit --no-coverage --filter FirewallCheckCommandTest
```

Expected `OK (18 tests, 66 assertions)`. They drive the real script via `proc_open` rather than including it, because the exit code and the stdout/stderr split are both part of the contract and neither can be asserted honestly in-process.

```bash
composer test    # 863 unit (+18) + 49 integration
```

## 2. By hand

Save this as `firewall.yml`. It needs no external databases and no API keys — in-memory storage, and CRS ships with the package — so it runs anywhere the library installs.

```yaml
# firewall.yml — a small config for exercising bin/firewall-check.
global:
  mode: block
  banning_message: "Request {{request.id}} blocked."

storage:
  type: "Kanopi\\Firewall\\Storage\\InMemoryStorage"

plugins:
  # Trusted addresses bypass everything below.
  - plugin: "Kanopi\\Firewall\\Plugins\\IpAddress"
    response: allow
    weight: -200
    enable: true
    config:
      - 198.51.100.10

  # Known-bad addresses.
  - plugin: "Kanopi\\Firewall\\Plugins\\IpAddress"
    response: block
    weight: -100
    enable: true
    config:
      - 203.0.113.0/24

  # Paths that should never be reachable.
  - plugin: "Kanopi\\Firewall\\Plugins\\Url"
    response: block
    weight: -50
    enable: true
    config:
      - "path@starts_with:/wp-admin"
      - "path:/.env"

  # Scanners and crawlers.
  - plugin: "Kanopi\\Firewall\\Plugins\\UserAgent"
    response: block
    weight: -10
    enable: true
    config:
      - "bot:true"
      - "client.name@contains:sqlmap"

  # OWASP Core Rule Set. No external data needed.
  - plugin: "Kanopi\\Firewall\\Plugins\\Crs"
    response: block
    weight: 50
    enable: true
    config:
      paranoia: 1
      mode: block
```

Every line below was run against exactly that file; the exit code is in the comment.

```bash
php bin/firewall-check --config=firewall.yml --ip=1.1.1.1 --url=/                      # ALLOWED     0
php bin/firewall-check --config=firewall.yml --ip=203.0.113.5 --url=/                  # IP Address  1
php bin/firewall-check --config=firewall.yml --ip=1.1.1.1 --url=/wp-admin/             # URL         1
php bin/firewall-check --config=firewall.yml --ip=1.1.1.1 --url=/.env                  # URL         1
php bin/firewall-check --config=firewall.yml --header='User-Agent: sqlmap/1.8'         # User Agent  1
php bin/firewall-check --config=firewall.yml --header='User-Agent: Googlebot/2.1'      # User Agent  1
php bin/firewall-check --config=firewall.yml --url="/search?q=1'+UNION+SELECT+1,2,3--" # CRS         1

# The allow bucket runs first, so a trusted address is let through a path
# that would otherwise be blocked:
php bin/firewall-check --config=firewall.yml --ip=198.51.100.10 --url=/wp-admin/       # ALLOWED     0
```

### `--explain` on a request that reaches the end of the chain

```console
$ php bin/firewall-check --config=firewall.yml --url="/search?q=1'+UNION+SELECT+1,2,3--" --explain

BLOCKED  GET /search?q=1'+UNION+SELECT+1,2,3--
  client            127.0.0.1
  blocked by        CRS
  status            403
  reason            Request 63284AE5F8F8F2FC006ACC7E1BD32A16 blocked.
  storage           throwaway (in-memory)

Plugins evaluated, in order:
  pass    IP Address                     0.05 ms
  pass    IP Address                     0.05 ms
  pass    URL                            0.30 ms
  pass    User Agent                   659.65 ms
  MATCH   CRS                           29.91 ms
```

Both `IpAddress` entries appear because the allow and block buckets each hold one. The timings are real: `matomo/device-detector` dominates on a cold call, which lines up with what the #100 benchmark found.

### `--explain` when something short-circuits

```console
$ php bin/firewall-check --config=firewall.yml --ip=203.0.113.5 --url=/wp-admin/ --explain

BLOCKED  GET /wp-admin/
  client            203.0.113.5
  blocked by        IP Address
  status            400
  ...

Plugins evaluated, in order:
  pass    IP Address                     0.05 ms
  MATCH   IP Address                     0.06 ms

Configured but not reached:
  block     Kanopi\Firewall\Plugins\Url              weight -50
  block     Kanopi\Firewall\Plugins\UserAgent        weight -10
  block     Kanopi\Firewall\Plugins\Crs              weight 50
```

If you were testing whether the URL rules catch `/wp-admin/`, this says the test never reached them — the IP list matched first.

### `--json`

```console
$ php bin/firewall-check --config=firewall.yml --ip=203.0.113.5 --url=/wp-admin/ --json
{
    "verdict": "blocked",
    "plugin": "IP Address",
    "status": 400,
    "reason": "Request 77D0FB0963139578FCAA37E31F35FCF2 blocked.",
    "request": {
        "ip": "203.0.113.5",
        "method": "GET",
        "path": "/wp-admin/",
        "query": null
    },
    "storage": "throwaway (in-memory)"
}
```

## 3. Verify the safety property

The most important behaviour. Point it at a config using `FileStorage` and confirm the file is never created:

```bash
cat > /tmp/fw-live.yml <<'YAML'
global: { mode: block }
storage:
  type: 'Kanopi\Firewall\Storage\FileStorage'
  config: { storage_file: /tmp/must-not-exist.json }
plugins:
  - plugin: "Kanopi\\Firewall\\Plugins\\IpAddress"
    response: block
    enable: true
    config: [203.0.113.0/24]
YAML

rm -f /tmp/must-not-exist.json
php bin/firewall-check --config=/tmp/fw-live.yml --ip=203.0.113.5   # BLOCKED
test -f /tmp/must-not-exist.json && echo "LEAKED (bug)" || echo "clean — storage untouched"

# And confirm the opt-in really does write:
php bin/firewall-check --config=/tmp/fw-live.yml --ip=203.0.113.5 --live-storage
test -f /tmp/must-not-exist.json && echo "written, as documented"
```

## 4. Confirm the CLI-bypass trap is actually closed

If mode-forcing ever regresses, everything reports allowed. This is the canary:

```bash
  --filter testBlockedRequestIsNotSilentlyAllowedInCli
```

## 5. Docs

```bash
mkdocs build --strict
```

New page: **Guides → Checking a Request**.

## Note for reviewers

`check:code` and `check:security` target `src/` only, so `bin/firewall-check` sits outside the standard gates. I ran both against it manually and it is clean, including **PHPStan at level max** — that run is what surfaced a real bug where `getopt()` returns an array for a repeated flag and `(string)` cast emitted "Array to string conversion" (`--ip=a --ip=b` now takes the last value). Worth a closer read than usual since CI is not backing you up on this file.

Whether these paths should be added to the `check:*` scripts is a reasonable follow-up, but it would pull `tests/Performance/bin/*` in too and is a bigger change than this PR should make.

## Follow-up

#104 (now merged) added `find()` / `deleteMatching()` for "who is blocked?"; this answers "would this be blocked, and by what?". This branch was cut before that landed, so the guide deliberately does **not** link to the Searching and Un-blocking section — `mkdocs --strict` correctly rejected the dangling anchor at the time. Worth adding that cross-link in a follow-up now the anchor exists.

### DIFF
--- a/bin/firewall-check
+++ b/bin/firewall-check
@@ -1,0 +1,507 @@
+#!/usr/bin/env php
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Firewall package.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/*
+ * Ask whether a given request would be blocked, and by what (#105).
+ *
+ * Everything this needs already existed — `Firewall::evaluate()` takes a
+ * Request, and `PluginManager::evaluate()` already records every plugin's
+ * result and timing. What did not exist was a correct way to call it from a
+ * terminal, because assembling it by hand has three edges that fail *quietly*:
+ *
+ *   1. `evaluate()` returns TRUE immediately when PHP_SAPI is 'cli' unless the
+ *      mode is `exception`. Forget that and every request reports "allowed",
+ *      including ones that plainly block. A checker that always says "nothing
+ *      is blocked" is worse than no checker.
+ *
+ *   2. Overrides are PropertyAccess bracket paths, not nested arrays.
+ *      `['global' => ['mode' => 'exception']]` is silently ignored and lands
+ *      you back in (1).
+ *
+ *   3. A block has side effects: it writes to storage, records an offense and
+ *      applies `blocking_escalation`. Pointed at a production config, a naive
+ *      "check" bans the address it was asked about.
+ *
+ * This script handles all three. Storage is swapped for a throwaway by
+ * default, so a check cannot ban anyone; pass --live-storage when you
+ * deliberately want the durable blocklist consulted.
+ */
+
+use Kanopi\Firewall\Exception\ChallengeRequiredException;
+use Kanopi\Firewall\Exception\ChallengeSolvedException;
+use Kanopi\Firewall\Exception\FirewallBlockedException;
+use Kanopi\Firewall\Firewall;
+use Kanopi\Firewall\Storage\InMemoryStorage;
+use Kanopi\Firewall\Utility\PluginConfigNormalizer;
+use Monolog\Handler\TestHandler;
+use Monolog\Level;
+use Symfony\Component\HttpFoundation\Request;
+
+// Exit codes, chosen so the tool composes in scripts and CI. The verdict is
+// the exit status, which is why "blocked" is not lumped in with "error".
+const EXIT_ALLOWED = 0;
+const EXIT_BLOCKED = 1;
+const EXIT_CHALLENGED = 2;
+const EXIT_USAGE = 64;
+const EXIT_INTERNAL = 70;
+
+// Works both in this repository and when installed as a dependency, where the
+// script is symlinked into the consumer's vendor/bin.
+$autoloaders = [
+    __DIR__ . '/../vendor/autoload.php',
+    __DIR__ . '/../../../autoload.php',
+];
+
+$loaded = false;
+
+foreach ($autoloaders as $autoloader) {
+    if (is_file($autoloader)) {
+        require $autoloader;
+        $loaded = true;
+        break;
+    }
+}
+
+if (!$loaded) {
+    fwrite(STDERR, "Could not locate the composer autoloader. Run `composer install`.\n");
+    exit(EXIT_INTERNAL);
+}
+
+$options = getopt('h', [
+    'config:',
+    'ip::',
+    'url::',
+    'method::',
+    'header:',
+    'body::',
+    'explain',
+    'json',
+    'live-storage',
+    'help',
+]);
+
+if ($options === false || isset($options['help']) || isset($options['h'])) {
+    usage();
+    exit($options === false ? EXIT_USAGE : EXIT_ALLOWED);
+}
+
+/**
+ * Print usage.
+ */
+function usage(): void
+{
+    $script = basename(__FILE__);
+
+    echo <<<TXT
+    Check whether a request would be blocked by a firewall configuration.
+
+    USAGE
+      {$script} --config=FILE [options]
+
+    REQUIRED
+      --config=FILE      Firewall config file. Repeatable; merged in order,
+                         exactly as Firewall::create() merges them.
+
+    REQUEST
+      --ip=ADDRESS       Client IP (IPv4 or IPv6). Default 127.0.0.1
+      --url=URL          Path, with optional query string. Default /
+      --method=VERB      HTTP method. Default GET, or POST when --body is given
+      --header=NAME:VAL  Request header. Repeatable
+      --body=STRING      Request body
+
+    OUTPUT
+      --explain          Show every plugin that evaluated, with result and
+                         timing, plus the plugins that never ran
+      --json             Machine-readable output
+
+    BEHAVIOUR
+      --live-storage     Use the configured storage backend instead of a
+                         throwaway. The durable blocklist is then consulted —
+                         but a block will also be RECORDED, which can ban the
+                         address you are asking about. Off by default.
+
+    EXIT CODES
+      0  allowed          64  usage error
+      1  blocked          70  internal error
+      2  challenged
+
+    EXAMPLES
+      {$script} --config=firewall.yml --ip=203.0.113.5 --url=/wp-admin/
+      {$script} --config=firewall.yml --url='/search?q=1%27+UNION+SELECT' --explain
+      {$script} --config=firewall.yml --header='User-Agent: sqlmap/1.8' --json
+
+    TXT;
+}
+
+/**
+ * Normalise a repeatable option into a list.
+ *
+ * getopt() hands back a string for one occurrence and an array for several,
+ * which is a reliable source of "works with two headers, breaks with one".
+ *
+ * @param array<string, mixed> $options
+ * @param string $name
+ *
+ * @return array<int, string>
+ */
+function repeatable(array $options, string $name): array
+{
+    if (!isset($options[$name])) {
+        return [];
+    }
+
+    $value = $options[$name];
+
+    return array_values(array_map('strval', is_array($value) ? $value : [$value]));
+}
+
+/**
+ * Read a single-valued option.
+ *
+ * getopt() hands back an array when a flag is repeated, so a bare
+ * `(string) $options['ip']` emits "Array to string conversion" the moment
+ * someone passes `--ip` twice. Last occurrence wins, which is what a reader
+ * would assume from the command line they typed.
+ *
+ * @param array<string, mixed> $options
+ * @param string $name
+ * @param string|null $default
+ *
+ * @return string|null
+ */
+function single(array $options, string $name, ?string $default = null): ?string
+{
+    if (!array_key_exists($name, $options)) {
+        return $default;
+    }
+
+    $value = $options[$name];
+
+    if (is_array($value)) {
+        $value = end($value);
+    }
+
+    return is_scalar($value) ? (string) $value : $default;
+}
+
+$configs = repeatable($options, 'config');
+
+if ($configs === []) {
+    fwrite(STDERR, "Error: at least one --config=FILE is required.\n\n");
+    usage();
+    exit(EXIT_USAGE);
+}
+
+foreach ($configs as $config) {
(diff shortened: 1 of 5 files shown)


kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/ea4bbd0e-37e9-488e-b1ac-3eccf86ce39d/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/ea4bbd0e-37e9-488e-b1ac-3eccf86ce39d/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 94.27
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/31352c75-b119-4980-9fa7-9d4c0b12b137/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 94.30
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/efa0b730-c7ae-4c34-b5fe-aac488adf6d8/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 94.30
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/01590860-3ae3-4019-a78c-1f1ecad6cfb3/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 94.30
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/eb274583-9649-416f-b72f-3539d93a4fb5/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 94.27
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/0322f037-ac0c-47a9-a59e-3e98bbdb5bca/artifacts/0/reports/index.html)

Issue #109, opened 2026-07-30
bot:true misses sqlmap, nikto, curl and other tooling (3.x: reconsider the bot detection source)


sean-e-dietrich:
## Summary

`bot:true` is documented in `docs/plugins/user-agent.md` and is one of the first rules people reach for. It is backed by `DeviceDetector::isBot()`, which **misses roughly half the attack tooling a firewall exists to stop** — including sqlmap and Nikto.

This is a detection-quality gap, not a performance one. Fixing it changes what `bot:true` matches, so it belongs in 3.x. The performance work on the same plugin (#107, #108) is independent and stays in 2.x.

## Evidence

`isBot()` against a representative set, with `jaybizzle/crawler-detect` alongside for comparison:

| User agent | `DeviceDetector::isBot()` | `CrawlerDetect::isCrawler()` |
|---|---|---|
| **sqlmap** | **miss** | catch |
| **nikto** | **miss** | catch |
| **curl** | **miss** | catch |
| **python-requests** | **miss** | catch |
| **Go-http-client** | **miss** | catch |
| masscan | catch | catch |
| nmap | catch | catch |
| zgrab | catch | catch |
| wpscan | catch | catch |
| nuclei | catch | catch |
| dirbuster | catch | catch |
| googlebot | catch | catch |
| bingbot | catch | catch |
| ahrefs | catch | catch |
| gptbot | catch | catch |
| Chrome desktop | pass | pass |
| Safari iOS | pass | pass |

Neither library false-positives on a real browser, so this is a pure coverage difference: **5 of 11 scanner / tooling agents slip past `bot:true` today.**

The practical consequence is that an operator who writes

```yaml
- plugin: "Kanopi\\Firewall\\Plugins\\UserAgent"
  response: block
  config:
    - "bot:true"
```

reasonably believes they have blocked scanners, and sqlmap walks straight through. The sample config in #106 only catches it because it *also* carries `client.name@contains:sqlmap` — a workaround for a gap most users will not know to work around.

## The nuance that makes this a design decision, not a swap

`crawler-detect` is not simply "better". The two libraries encode different definitions:

- **`DeviceDetector::isBot()`** means *"a known bot in our curated database"* — with name, category and producer attached.
- **`CrawlerDetect::isCrawler()`** is broader, and deliberately includes generic HTTP client libraries: `curl`, `python-requests`, `Go-http-client`.

[...]

Issue #108, opened 2026-07-30
UserAgent: only run the device-detector parse phases the rules actually need


sean-e-dietrich:
## Summary

`UserAgent::evaluate()` runs a full `DeviceDetector::parse()` on every request regardless of what the configured rules actually ask for. Device parsing — brand and model detection — is **~85% of the cost**, and most configurations never reference it.

A config that only uses `bot:` and `client.*` could skip that phase entirely: roughly **630 ms → 115 ms** on a cold parse, independent of and multiplicative with the caching in #107.

## Measurement

`DeviceDetector::parse()` (`vendor/matomo/device-detector/DeviceDetector.php:595`) runs four phases in order, returning early if the UA is a bot:

```php
$this->parseBot();
if ($this->isBot()) { return; }
$this->parseOs();
// clients …
// device …
```

Timing each phase individually:

| Phase | iphone safari | sqlmap |
|---|---|---|
| `parseBot()` | 48.5 ms | 46.1 ms |
| `parseOs()` | 16.2 ms | 16.9 ms |
| `parseClient()` | 50.5 ms | 54.4 ms |
| **`parseDevice()`** | **515.6 ms** | **540.2 ms** |

`parseDevice()` walks the brand/model corpus, which is by far the largest part of the regex set. It is also why an ordinary mobile browser costs ~596 ms while Googlebot costs ~48 ms — the bot short-circuits before ever reaching it. See #107 for the per-UA breakdown.

## The opportunity

`src/Plugins/UserAgent.php:50-52` parses unconditionally:

```php
public function evaluate(Request $request): bool
{
    $userAgent = $request->headers->get('User-Agent', '');
    $this->deviceDetector = $this->detectDevice($userAgent);
```

But `getValue()` (`src/Plugins/UserAgent.php:127-150`) shows each config variable needs only one phase:

| Config variable | Phase required |
|---|---|
| `bot`, `bot.name`, `bot.category`, `bot.producer` | `parseBot()` |
| `os.name`, `os.short_name`, `os.version` | `parseOs()` |
| `client.name`, `client.type`, `client.version` | `parseClient()` |
| `device.type`, `brand`, `model` | **`parseDevice()`** |

The sample config in #106 uses `bot:true` and `client.name@contains:sqlmap`. Neither needs `parseDevice()`, so that config is paying ~515 ms per process for a phase whose output it never reads.

[...]

PR #110, opened 2026-07-30
Cache the device-detector regex corpus (Fixes #107)

### LINKED ISSUE #107: UserAgent: cache the device-detector regex corpus (~630ms per process)
## Summary

`UserAgent` pays roughly **630 ms per PHP process** re-compiling `matomo/device-detector`'s regex corpus, because nothing ever calls `DeviceDetector::setCache()`. Attaching a persistent cache takes that to **~45 ms**, measured across separate processes.

On current evidence this is the single largest available win in the plugin layer — larger than the mmdb caching #6 proposes.

## Measurements

`src/Plugins/UserAgent.php:83-91` constructs a fresh `DeviceDetector` per evaluation and calls `->parse()`. The corpus is 20 YAML files totalling 1.7 MB, and it is compiled on first parse.

Within a single process:

| | Time |
|---|---|
| 1st parse (cold) | **630.25 ms** |
| 2nd parse | 4.48 ms |
| 3rd parse | 3.55 ms |
| 4th parse | 4.17 ms |
| Peak memory | 20 MB |

So ~626 ms of one-time corpus loading, then ~4 ms per parse.

With a `PSR6Bridge` over a `FilesystemAdapter`, measured in **separate processes** so the cache is genuinely cold/warm rather than warm in-process:

| | Time |
|---|---|
| Cold, populating the cache | 634.80 ms |
| Warm, new process | **45.43 ms** |
| Warm, new process | **44.94 ms** |

## Cost depends on the user agent, and ordinary traffic is the worst case

The original framing of this issue implied scanners were the problem. That was wrong. Cold cost per UA:

| User agent | cold | `isBot()` |
|---|---|---|
| googlebot | **48.08 ms** | true |
| chrome desktop | 109.93 ms | false |
| **iphone safari** | **596.11 ms** | false |
| sqlmap | 627.76 ms | false |
| unrecognised string | 637.19 ms | false |
| empty | 0.00 ms | false |

Bots are the *cheap* case, because `parse()` returns early once `parseBot()` matches. The expensive cases are **ordinary mobile browsers** and anything unrecognised — that is normal production traffic, not an edge case.

Two things worth noting while here:

- `device-detector` reports **`isBot() === false` for sqlmap** — and for nikto, curl, python-requests and Go-http-client. That is a detection-quality gap rather than a performance one, tracked separately for 3.x in #109; the sample config in #106 only catches sqlmap because it also carries `client.name@contains:sqlmap`.
- The empty-UA case is free, because `parse()` short-circuits on a user agent containing no letters.

## Where this shows up

Three independent observations, all the same root cause:

- **`bin/firewall-check`** (#105): one process, one request, so it pays the full cost — `--explain` reports `User Agent  639.65 ms` against `CRS  29.91 ms` and `URL  0.30 ms`.
- **The #100 benchmark**: the `user-agent` scenario measured ~1,095 ms p95 and **18 MB peak memory against 2 MB** for every other scenario. The harness recreates the container per scenario, so all 64 workers start cold and a large share of requests are some worker's first.
- **PHP-FPM in production**: each worker pays the cold cost on its first request, then ~4 ms. With `pm.max_requests` set — 500 is a common default — workers recycle and re-pay it periodically.

The 20 MB peak measured here matches the 18 MB the benchmark reported, which is corroboration that these are the same phenomenon rather than two separate ones.

## Should we just use a different library?

Asked and answered: **no.** Benchmarked against the plausible alternatives on the same machine:

| Library | cold | warm | peak | last release |
|---|---|---|---|---|
| matomo/device-detector | 110–637 ms | 1.38 ms | 6 MB | 2 months ago |
| whichbrowser/parser | 18.15 ms | 0.49 ms | 6 MB | **2 years ago** |
| ua-parser/uap-php | 10.31 ms | 0.31 ms | 4 MB | 1 year ago |
| donatj/phpuseragentparser | 0.18 ms | 0.017 ms | 2 MB | 1 month ago |
| jaybizzle/crawler-detect | 3.01 ms | 0.018 ms | 2 MB | 2 weeks ago |

They are genuinely faster, and none of them can replace it. Coverage against the variables `docs/plugins/user-agent.md` documents as public config API:

| Plugin variable | device-detector | whichbrowser | uap-php | donatj |
|---|---|---|---|---|
| `bot:true` | yes | `device.type==='bot'` | **no API** | no |
| `bot.name` / `.category` / `.producer` | yes | no | no | no |
| `device.type` | `smartphone` | `mobile` | no | no |
| `client.name` | `Chrome Mobile` | `Chrome` | `Chrome Mobile` | `Chrome` |
| `client.type` | `browser` | no | no | no |
| `os.short_name` | `AND` | no | no | no |
| `brand` | `Google` | empty | `Google` | no |
| `model` | `Pixel 8` | `Pixel 8` | `Pixel 8` | no |

The rows where a value merely *differs* are more dangerous than the ones that are missing. A missing field breaks a config loudly; `device.type` returning `mobile` instead of `smartphone`, or `client.name` returning `Chrome` instead of `Chrome Mobile`, makes an existing rule **silently stop matching** on a blocking security component.

Add that `whichbrowser` has not released in two years, and that device-detector carries by far the richest bot database — name, category, producer — which is exactly what a firewall wants, and switching is clearly the wrong trade.

## Proposal

Make the cache configurable on the plugin, and attach it in `detectDevice()`:

```yaml
- plugin: "Kanopi\\Firewall\\Plugins\\UserAgent"
  response: block
  enable: true
  metadata:
    cache:
      adaptor: "Symfony\\Component\\Cache\\Adapter\\FilesystemAdapter"
      args: ['device-detector', 0, '/var/cache/firewall']
  config:
    - "bot:true"
```

Two conventions already exist in this repository and this should follow them rather than invent a third:

- **`CacheRateLimitStorage`** (`src/RateLimitStorage/CacheRateLimitStorage.php:38`) takes `['adaptor' => class-string|CacheItemPoolInterface, 'args' => array, 'ttl' => int]`, accepts either a class name or an already-constructed instance, and logs a warning instead of throwing when the adaptor cannot be resolved. Same shape works here.
- **`AbuseIpdb`** establishes the precedent for a default cache location: an explicit `cache_dir`, else the `KANOPI_FIREWALL_CACHE_DIR` constant, else the system temp directory.

`symfony/cache` is already a hard dependency and `device-detector` ships `Cache/PSR6Bridge`, so no new packages are required.

## Cache invalidation is already handled

Worth stating up front, because it is the obvious objection. `device-detector` namespaces its own cache keys by version:

```php
// vendor/matomo/device-detector/Parser/AbstractParser.php:284
$cacheKey = 'DeviceDetector-' . DeviceDetector::VERSION . 'regexes-' . $this->getName();
```

So a `composer update` that bumps device-detector produces different keys and the old entries simply age out. There is no stale-regex risk to design around.

## Open questions

1. **Should caching be on by default?** Off by default means almost nobody gets the win, since this is invisible unless you profile. On by default needs a writable directory, and the `AbuseIpdb` precedent (temp dir fallback, warn and continue if unwritable) suggests that is acceptable. Leaning on by default with a documented opt-out.

2. **Which adaptor as the default?** `FilesystemAdapter` works everywhere and gave the 45 ms figure above. APCu would be faster still but is not always installed. Filesystem seems the right default with APCu documented as an upgrade.

3. **Does `StaticCache` help anything?** `device-detector` ships an in-process `Cache/StaticCache`. It would not survive the process, so it does nothing for PHP-FPM's first-request-per-worker cost — but it would help any long-running consumer (a queue worker, a CLI batch) that evaluates many requests. Cheap enough to consider as a fallback when no adaptor is configured.

## Relationship to the other UserAgent issue

Caching is one of two independent wins, and this issue is deliberately scoped to just this one — attach a cache, keep the parse identical. The other is skipping parse phases the configured rules never reference, which is a larger win but needs design work; it is tracked separately in #108 so this one can land without waiting.

Do this one first: it is a `setCache()` call plus a config knob, with no behavioural change to the parse itself.

## Relationship to #6

#6 (plugin caching) proposes caching per-IP results on the premise that *"the cost to read the mmdb can be taxing"* for `GeoLocation` / `Asn`. That premise is unverified and, on this evidence, likely mis-aimed: mmdb lookups are b-tree descents into a memory-mapped file, whereas the measured cost here is corpus compilation.

This issue is deliberately narrow and independent — a `setCache()` call plus a config knob, no new caching layer, no invalidation semantics to design. It should not wait on #6, and #6 should probably be re-scoped once the nightly benchmark reports real `geolocation` and `asn` numbers.

## Verifying a fix

```bash
# Before and after, on the sample config from #106:
bin/firewall-check --config=firewall.yml --header='User-Agent: sqlmap/1.8' --explain
```

The `User Agent` row in the `Plugins evaluated` table is the number to watch. The #100 benchmark's `user-agent` scenario gives the same signal under load, and its `peak_kb` column should fall towards the ~2 MB the other scenarios report.

sean-e-dietrich:
Fixes #107.

`matomo/device-detector` compiles 20 YAML files totalling 1.7 MB on the first parse in each PHP process. Nothing ever called `setCache()`, so **every process paid it**.

Measured end to end through `bin/firewall-check` against an iPhone Safari agent:

```
618 ms   first request (populating the cache)
 23 ms   every subsequent process
618 ms   with caching disabled, every time
```

Under PHP-FPM each worker paid that on its first request, and again after every `pm.max_requests` recycle.

## The cost is not a scanner problem

This issue originally assumed scanners were the expensive case. They are the **cheap** one — `parse()` returns early once `parseBot()` matches:

| User agent | cold | `isBot()` |
|---|---|---|
| googlebot | **48 ms** | true |
| chrome desktop | 110 ms | false |
| **iphone safari** | **596 ms** | false |
| sqlmap | 628 ms | false |
| unrecognised string | 637 ms | false |

Ordinary mobile browsers are among the worst cases, because brand and model detection walks the largest part of the corpus. This hits real traffic hardest, which makes it more worth fixing than the original framing suggested.

It also explains the #100 benchmark's `user-agent` scenario: ~1,095 ms p95 and 18 MB peak against 2 MB elsewhere. The harness recreates the container per scenario, so all 64 workers start cold and a large share of requests are some worker's first.

## Configuration

On by default, no configuration required. `metadata.cache` accepts:

| Value | Effect |
|---|---|
| *(omitted)* | Filesystem cache in `KANOPI_FIREWALL_CACHE_DIR`, else the system temp directory |
| `false` | Caching off |
| `{dir: /path}` | Same default backend, different location |
| `{adaptor: FQCN, args: [...]}` | Any PSR-6 pool |
| an object | An already-constructed pool, injected via config overrides |

The `['adaptor', 'args']` shape is the one `CacheRateLimitStorage` already accepts, so there is one convention rather than two. The temp-directory default follows what `AbuseIpdb` already does for its verdict cache.

**On by default** because off-unless-configured would mean almost nobody benefits — the cost is invisible without profiling.

## Nothing here can take a site down

- An unusable adaptor (missing class, or a class that is not a PSR-6 pool) falls back to the default and warns.
- An unwritable directory logs at `warning` and continues **uncached**.

A plugin that refused to evaluate because a cache directory was read-only would turn an optimisation into an outage.

## Cache invalidation needs no handling

`device-detector` keys its entries by `DeviceDetector::VERSION` (`Parser/AbstractParser.php:284`), so a `composer update` that bumps the package produces new keys and the stale ones age out. Nothing to clear by hand.

---

# How to test

## 1. Automated

```bash
vendor/bin/phpunit -c phpunit.xml --testsuite unit --no-coverage --filter UserAgentCacheTest
```

Expected `OK (19 tests, 34 assertions)`.

The wiring tests are the boring ones. **The tests that matter assert caching does not change detection** — cold and warm verdicts compared against an uncached run across `bot`, `device.type`, `os.name`, `client.name` and `bot.name`. A cache that quietly altered a verdict would be a security regression dressed as an optimisation.

```bash
composer test    # 931 unit + 49 integration
```

## 2. See the improvement yourself

```bash
cat > /tmp/ua.yml <<'YAML'
global: { mode: block }
storage: { type: 'Kanopi\Firewall\Storage\InMemoryStorage' }
plugins:
  - plugin: "Kanopi\\Firewall\\Plugins\\UserAgent"
    response: block
    enable: true
    config:
      - "client.name@contains:sqlmap"
YAML

UA='User-Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 17_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Mobile/15E148 Safari/604.1'

# Clear the cache. NOTE: sys_get_temp_dir() is not /tmp on macOS.
rm -rf "$(php -r 'echo sys_get_temp_dir();')/kanopi-firewall-device-detector"

bin/firewall-check --config=/tmp/ua.yml --header="$UA" --explain | grep 'User Agent'   # ~618 ms, cold
bin/firewall-check --config=/tmp/ua.yml --header="$UA" --explain | grep 'User Agent'   # ~23 ms
bin/firewall-check --config=/tmp/ua.yml --header="$UA" --explain | grep 'User Agent'   # ~23 ms
```

Then add `metadata: {cache: false}` to the plugin entry and confirm it returns to ~618 ms on every run.

> Worth flagging because it cost me a wrong measurement: on macOS `sys_get_temp_dir()` is `/var/folders/…/T`, not `/tmp`. Clearing `/tmp/kanopi-firewall-device-detector` silently clears nothing, and a "cold" run then reports ~40 ms because the real cache was still warm.

## 3. Confirm it degrades safely

```bash
# Unwritable cache location — must still evaluate, just slowly.
  --filter 'testUnwritableCacheDirectoryDoesNotBreakEvaluation|testUnusableAdaptorFallsBackAndWarns|testNonPoolClassIsRejected'
```

## 4. Docs

```bash
mkdocs build --strict
```

New **Caching** section in `docs/plugins/user-agent.md`.

## Follow-ups, unchanged by this

- **#108** — skipping the parse phases the configured rules never reference. Independent: this removes the corpus *compile*, that removes the corpus *walk*. `parseDevice()` is ~85% of the remaining cost, so the two compound.
- **#109** — 3.x, the `bot:true` coverage gap. Detection quality, not speed.

### DIFF
--- a/docs/plugins/user-agent.md
+++ b/docs/plugins/user-agent.md
@@ -42,6 +42,67 @@ plugins:
           - "client.version@less_than:80"
 ```
 
+## Caching
+
+The plugin's detection is backed by `matomo/device-detector`, which compiles a 1.7&nbsp;MB corpus of regex files on the first parse in each PHP process. That costs **110–637&nbsp;ms** depending on the user agent — ordinary mobile browsers are among the worst cases, because brand and model detection walks the largest part of the corpus. Once warm it is roughly 4&nbsp;ms.
+
+Under PHP-FPM every worker pays that on its first request, and again after each `pm.max_requests` recycle. The plugin therefore caches the compiled corpus **by default**:
+
+```
+618 ms   first request (populating the cache)
+ 23 ms   every subsequent process
+618 ms   with caching disabled, every time
+```
+
+No configuration is needed. The cache is written to `KANOPI_FIREWALL_CACHE_DIR` when that constant is defined, otherwise to a `kanopi-firewall-device-detector` directory inside the system temp directory — the same convention the [AbuseIPDB plugin](abuseipdb.md) uses for its verdict cache.
+
+### Pointing it somewhere else
+
+```yaml
+- plugin: "Kanopi\\Firewall\\Plugins\\UserAgent"
+  response: block
+  enable: true
+  metadata:
+    cache:
+      dir: /var/cache/firewall
+  config:
+    - "bot:true"
+```
+
+### Using a different backend
+
+Any PSR-6 pool works. The shape matches the one [rate limiting](rate-limit.md) already accepts, so there is a single convention to learn:
+
+```yaml
+    metadata:
+      cache:
+        adaptor: "Symfony\\Component\\Cache\\Adapter\\ApcuAdapter"
+        args: ['device-detector', 0]
+```
+
+An already-constructed pool can be injected through [configuration overrides](../configuration/overrides.md), since YAML cannot carry an object:
+
+```php
+Firewall::create([__DIR__ . '/firewall.yml'], [
+    '[plugins][0][metadata][cache][adaptor]' => $myCachePool,
+]);
+```
+
+### Turning it off
+
+```yaml
+    metadata:
+      cache: false
+```
+
+Detection is unchanged either way — only the speed differs.
+
+!!! note "Upgrades invalidate the cache automatically"
+
+    `device-detector` keys its cache entries by its own version, so a `composer update` that bumps the package produces new keys and the stale entries simply age out. There is nothing to clear by hand.
+
+A cache that cannot be created never stops the plugin working: the failure is logged at `warning` and detection continues uncached. An optimisation should not be able to take a site down.
+
 ## Available Variables
 
 - `bot` - Whether the user agent is a bot ("true" or "false")
--- a/src/Plugins/UserAgent.php
+++ b/src/Plugins/UserAgent.php
@@ -11,10 +11,14 @@ declare(strict_types=1);
 
 namespace Kanopi\Firewall\Plugins;
 
+use DeviceDetector\Cache\CacheInterface;
+use DeviceDetector\Cache\PSR6Bridge;
 use DeviceDetector\ClientHints;
 use DeviceDetector\DeviceDetector;
 use DeviceDetector\Parser\Device\AbstractDeviceParser;
 use Kanopi\Firewall\Traits\EvaluateTrait;
+use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
 use Symfony\Component\HttpFoundation\Request;
 
 /**
@@ -28,6 +32,24 @@ class UserAgent extends AbstractPluginBase
      * Device Detector for the current request.
      */
     protected DeviceDetector $deviceDetector;
+
+    /**
+     * Regex-corpus cache handed to every DeviceDetector instance (#107).
+     *
+     * NULL when caching is switched off, or when the configured adaptor could
+     * not be built — in which case the plugin still works, just slowly.
+     */
+    protected ?CacheInterface $cache = null;
+
+    /**
+     * Whether the cache has been resolved yet.
+     *
+     * Resolution is deferred to the first evaluation rather than done in the
+     * constructor: plugins are instantiated lazily by `PluginManager`, but a
+     * config can still declare a plugin that never runs, and building a cache
+     * adaptor is not free — it creates directories.
+     */
+    private bool $cacheResolved = false;
 
     /**
      * {@inheritdoc}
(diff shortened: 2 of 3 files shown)


kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 93.71
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/eb7eb50b-c010-4daf-8c05-e839911f48d7/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 93.71
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/dc9b9f03-970f-49b8-8850-9127e831a773/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 93.74
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/0ca4cd20-29b8-4e8f-a579-d7eb6faa0619/artifacts/0/reports/index.html)

PR #118, opened 2026-07-31
Add a %file(/path)% token for literal paths (Fixes #116)

### LINKED ISSUE #116: Add a %file(/path)% token for reading a file from a literal path
## Summary

There is no way to read a file into a config value from a **literal path**. Every token resolves through an environment variable: `TokenSubstitute` does

```php
$parts = \explode(':', $token);
$var   = \array_pop($parts);   // the last segment is always an env var name
```

and the value always originates from `getenv()` / `$_SERVER`.

A path *can* be smuggled in via `default:`, and it works:

```yaml
challenge:
  secret: '%env(file:default:/etc/firewall/hmac.key:UNUSED)%'
```

But that form is a workaround with two real hazards, and it is the only option anyone has today.

## Why the workaround is not good enough

**It depends on a variable never existing.** `UNUSED` has to stay undefined forever. If anything in the environment later defines it — a platform injecting variables, a `.env` file, a colleague reusing the name — the path silently becomes that value and the secret silently changes. Nothing in the config file signals that dependency.

**A colon truncates the path.** The token is `explode(':')`, so:

```
%env(file:default:/tmp/sec:rets/x.txt:UNUSED)%
  -> ConfigurationException: Path for file processor does not resolve: /tmp/sec
```

The codebase already knows about this class of problem — `raw_key:` exists specifically for keys containing `:` — but there is no equivalent escape for paths.

**It reads as a mistake.** A three-part chain ending in a variable that must not exist is not something a reviewer can look at and judge correct.

## Proposal

A first-class token that takes a literal path:

```yaml
challenge:
  secret: '%file(/etc/firewall/hmac.key)%'

global:
  banning_message: '%file(/etc/firewall/banned.html)%'
```

- **No colon parsing.** The token content is the whole path, so `:` is no longer special.
- **No phantom variable.** Nothing to keep undefined.
- **Same safety controls.** It should honour the existing `TokenSubstitute::enableUnsafeProcessors(['file'], [...])` opt-in and the base-directory allowlist, so nothing becomes *more* permissive than today.

## Implementation notes

`TokenSubstitute` matches tokens at three points, and a parallel form needs handling at each:

| Line | Purpose |
|---|---|
| `:209` | exact-match — the whole scalar is one token, returns a native type |
| `:215` | interpolation — a token inside a larger string |
| `:742` | resolving an `%env(...)%` used as an include path |

The existing pattern is `/^%env\(([^)]+)\)%$/`. Note `[^)]+`: a path containing `)` would break, so the new form should either accept a greedier match or document the limit.

## Open questions

1. **Should it require the `file` opt-in at all?** The risk `enableUnsafeProcessors()` guards against is *env-var injection* turning into arbitrary file reads. A literal path in a YAML file is only as trustworthy as that file — and if an attacker can edit your firewall config, the file read is not your problem. So an argument exists for `%file()%` needing no opt-in.

   Leaning towards **requiring it anyway**: consistent with `file:`, keeps one mental model, and the allowlist is cheap insurance for a security package. Worth deciding explicitly rather than by omission.

2. **Should `%require(...)%` follow?** Probably not. Executing a literal path is still executing a path, and the docs already advise preferring `file:` over `require:`. Adding the safer half only is defensible.

3. **Relative paths.** Should they resolve against the including config file's directory, as `configs:` entries do? That would be the consistent behaviour, and is likely what someone writing `%file(secrets/hmac.key)%` expects.

4. **Trailing whitespace.** A key file usually ends with a newline. `file:` returns contents verbatim today, so `%file()%` should too — but it is worth documenting, since an HMAC secret with a stray `\n` fails in a way that is annoying to diagnose.

## Related

#114 covers documenting the current state, and is worth doing regardless of whether this lands.

sean-e-dietrich:
Fixes #116.

Every route to a file's contents ran through an environment variable, because `$var` in an `%env(...)%` chain is always the last colon-separated segment:

```php
$parts = \explode(':', $token);
$var   = \array_pop($parts);
```

So a path known up front had exactly one option, and it was a workaround:

```yaml
secret: '%env(file:default:/etc/firewall/hmac.key:UNUSED)%'
```

Now:

```yaml
secret: '%file(/etc/firewall/hmac.key)%'
```

## The two hazards this removes

**A variable that must never exist.** The old form depends on `UNUSED` staying undefined forever. Define it — a platform injecting variables, a `.env` file, a colleague reusing the name — and the path silently becomes that value, with nothing in the config signalling the dependency.

**A colon truncates the path.** The chain is split on `:`, so `/tmp/sec:rets/key.txt` resolves as `/tmp/sec`.

Both are pinned by tests, so the improvement is measured against real behaviour rather than asserted:

- `testAPathContainingAColonWorks` — the new form
- `testTheOldFormStillTruncatesOnAColon` — the old one still fails, so the first test is measuring a genuine difference
- `testNoEnvironmentVariableCanRedirectTheRead` — sets `UNUSED=/etc/passwd` and confirms the read is unaffected

## Security posture is unchanged

Held to the same `enableUnsafeProcessors(['file'], [...])` opt-in and base-directory allowlist as the `file:` processor.

I considered exempting it — a literal path in a config file is only as trustworthy as that file, and an attacker who can edit your firewall config has better options than a file read. Decided against: one mental model is worth more than the convenience, and the allowlist still limits the blast radius if a config file is ever compromised. Raised as an open question in #116 and settled deliberately rather than by omission.

Covered by tests: refusal without the opt-in, a path outside the allowlist, and traversal out of it.

## Other decisions

- **Contents come back verbatim**, newline included, matching `file:`. Documented, since an HMAC secret carrying a stray `\n` fails in a way that is annoying to diagnose. The override route remains the place to `trim()`.
- **No `%require(...)%` counterpart.** Executing a literal path is still executing a path, and the docs already advise preferring `file:` over `require:`.
- **`%env(...)%` is untouched.** A test resolves both forms in the same string.

---

# How to test

## 1. Automated

```bash
vendor/bin/phpunit -c phpunit.xml --testsuite unit --no-coverage --filter FileTokenTest
```

Expected `OK (14 tests, 19 assertions)`.

```bash
composer test    # 1029 unit + 49 integration
```

## 2. By hand

```php
<?php
require 'vendor/autoload.php';
use Kanopi\Firewall\Utility\TokenSubstitute;

$dir = '/tmp/fw-file-token';
@mkdir($dir); @mkdir("$dir/sec:rets");
file_put_contents("$dir/hmac.key", "literal-secret\n");
file_put_contents("$dir/sec:rets/x.txt", 'colon-ok');

// Refused until you opt in.
try { TokenSubstitute::substitute("%file($dir/hmac.key)%"); }
catch (Throwable $e) { echo "refused: ", $e->getMessage(), "\n"; }

TokenSubstitute::enableUnsafeProcessors(['file'], [$dir]);

var_dump(TokenSubstitute::substitute("%file($dir/hmac.key)%"));        // "literal-secret\n"
var_dump(TokenSubstitute::substitute("%file($dir/sec:rets/x.txt)%"));  // "colon-ok"  <- colon works
var_dump(TokenSubstitute::substitute("key=[%file($dir/hmac.key)%]"));  // interpolates

// Outside the allowlist.
try { TokenSubstitute::substitute('%file(/etc/passwd)%'); }
catch (Throwable $e) { echo "blocked: ", $e->getMessage(), "\n"; }

// The old form, same colon path — still truncates.
try { TokenSubstitute::substitute("%env(file:default:$dir/sec:rets/x.txt:UNUSED)%"); }
catch (Throwable $e) { echo "old form: ", $e->getMessage(), "\n"; }
```

## 3. In a real config

```yaml
challenge:
  provider: math
  secret: '%file(/etc/firewall/hmac.key)%'
```

with `TokenSubstitute::enableUnsafeProcessors(['file'], ['/etc/firewall']);` before `Firewall::create()`.

## 4. Docs

```bash
mkdocs build --strict
```

New *Reading from a known path* section in `docs/configuration/environment-variables.md`, with a note explaining the old form for anyone who meets it in an existing config.

## Relationship to #117

#117 documents the workaround as the current state. This makes the workaround unnecessary. They touch the same section, so **whichever merges second will need a small rebase** — if this lands first, #117's *Reading from a known path* section collapses to the `!!! note` block already included here.

I kept them separate as asked, but they are worth reviewing together.

### DIFF
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -125,41 +125,49 @@ global:
 
 ### Reading from a known path
 
-When the path is known up front, you do not need an environment variable at all. `default:` supplies the path and `file:` reads it:
+When you already know the path, use `%file(...)%` rather than an environment variable:
 
 ```yaml
 challenge:
-  secret: '%env(file:default:/etc/firewall/hmac.key:UNUSED)%'
+  secret: '%file(/etc/firewall/hmac.key)%'
+
+global:
+  banning_message: '%file(/etc/firewall/banned.html)%'
 ```
 
-`UNUSED` is a variable name that must **never be defined**. The chain reads left to right: the variable is unset, so `default:` yields `/etc/firewall/hmac.key`, and `file:` reads that path.
+The token content is the whole path, so unlike the `file:` processor there is **no colon limitation** — `%file(/tmp/sec:rets/key.txt)%` works. There is also no environment variable involved, so nothing in the environment can redirect the read.
 
-This is the safer of the two forms — an attacker who can influence the environment cannot redirect a path that is written in the config file. Two caveats before you rely on it:
+The same opt-in and base-directory allowlist apply:
 
-!!! warning "The named variable must stay undefined"
+```php
+TokenSubstitute::enableUnsafeProcessors(['file'], ['/etc/firewall']);
+```
 
-    If anything ever defines `UNUSED` — a platform injecting variables, a `.env` file, a colleague reusing the name — the path silently becomes that value and the file read silently changes. Nothing in the config signals that dependency. Choose a name you are confident nothing else will use, and treat it as part of the configuration.
+A literal path in a config file is only as trustworthy as that file, so this form is lower risk than the env-var one — but it is held to the same controls so there is a single mental model, and the allowlist still limits the blast radius if a config file is ever compromised.
 
-    This is where the base-directory allowlist earns its keep. With `enableUnsafeProcessors(['file'], ['/etc/firewall'])` in place, a hijacked variable pointing at `/etc/passwd` fails loudly instead of being read:
+Contents come back **verbatim**, newline included, exactly as `file:` returns them. A key file written by an editor usually ends with `\n`, and an HMAC secret carrying a stray newline fails in a way that is annoying to diagnose — if that matters, read it through a [configuration override](overrides.md#reading-a-value-from-a-file) where you can `trim()` it.
+
+`%file(...)%` interpolates inside a larger string and resolves inside nested arrays, like `%env(...)%`.
+
+!!! note "The older workaround, and why to move off it"
+
+    Before this token existed, the only way to use a literal path was to chain `default:` into `file:`:
+
+    ```yaml
+    secret: '%env(file:default:/etc/firewall/hmac.key:UNUSED)%'
+    ```
+
+    It still works, and you will meet it in existing configs, but it carries two hazards that `%file(...)%` does not.
+
+    **`UNUSED` must never be defined by anything.** Define it — a platform injecting variables, a `.env` file, a colleague reusing the name — and the path silently becomes that value. Nothing in the config file signals the dependency. This is where the base-directory allowlist earns its keep: with `enableUnsafeProcessors(['file'], ['/etc/firewall'])` in place, a hijacked variable pointing at `/etc/passwd` fails loudly rather than being read:
 
     ```
     ConfigurationException: Path for file processor escapes the configured allowlist: /etc/passwd
     ```
 
-    Do not pass an empty allowlist when using this form.
+    Never pass an empty allowlist when using that form.
 
-!!! warning "A colon truncates the path"
-
-    Tokens are split on `:`, so a path containing one is cut short:
-
-    ```
-    %env(file:default:/tmp/sec:rets/key.txt:UNUSED)%
-      -> ConfigurationException: Path for file processor does not resolve: /tmp/sec
-    ```
-
-    This is the same limitation `raw_key:` exists to work around for keys, but there is no equivalent escape for paths. If your path contains a colon, use a [configuration override](overrides.md#reading-a-value-from-a-file) instead.
-
-If neither caveat sits well, reading the file in PHP and passing it as an [override](overrides.md#reading-a-value-from-a-file) avoids both.
+    **A colon truncates the path.** Tokens are split on `:`, so `%env(file:default:/tmp/sec:rets/key.txt:UNUSED)%` resolves the path as `/tmp/sec`. This is the same limitation `raw_key:` exists to work around for keys; there is no equivalent escape for paths.
 
 Prefer `file:` over `require:` whenever you can — reading a secret is far less dangerous than executing a path an attacker may influence.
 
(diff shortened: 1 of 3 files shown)


kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 93.69
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/df96d755-ff2c-4c49-8150-4a440b8529d3/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 93.72
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/64161158-399b-4885-9eb7-7063ed37c440/artifacts/0/reports/index.html)

kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/7ed8f788-a264-4014-a693-6925e283de91/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/7ed8f788-a264-4014-a693-6925e283de91/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 93.72
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/ecee4e52-afd1-47fd-9534-82dd7a6a18dd/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 93.72
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/2be04caa-ee89-44a4-b90b-4ff61b277427/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 93.69
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/f226ca72-8365-4f42-9db1-9e2457d52e47/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 93.72
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/cf1b1f49-0b4e-490a-a888-3f34b7b1dd5d/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 93.69
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/cdaa0d3c-6657-4dfc-8342-b563b9ea2dc2/artifacts/0/reports/index.html)

PR #126, opened 2026-08-04
Add a built-in Cloudflare Turnstile challenge provider (Fixes #125)

### LINKED ISSUE #125: Add a built-in Cloudflare Turnstile challenge provider
## Summary

Ship Cloudflare Turnstile as a built-in `challenge.provider`, alongside `math` and `altcha`.

Today the docs tell operators who want real bot resistance to write their own provider ([`docs/plugins/challenges.md`](../blob/2.x/docs/plugins/challenges.md) — "For stronger bot resistance (Turnstile, hCaptcha, reCAPTCHA, etc.), implement `ChallengeProviderInterface`"), and ship a partial Turnstile sketch as the worked example. That sketch leaves the hard parts — server-side verification, error handling, escaping, CSP — to every operator individually. `ChallengeProviderInterface`'s own docblock already names Turnstile as an expected implementation.

## Why Turnstile first

Free and unmetered, no per-request cost, no Google dependency, and far fewer GDPR complications than reCAPTCHA for EU clients. hCaptcha is a near-identical integration and a sensible follow-up; reCAPTCHA v2 also fits, while v3 does not (it is an invisible score with no user-completable step, so it belongs alongside `VulnerabilityScore` rather than in the challenge flow).

## What already works unchanged

The whole browser leg is provider-agnostic and needs no modification:

`interstitial → POST challenge.path → Firewall::isChallengeSubmission() → handleChallengeSubmission() → $provider->verifySolution() → mint pass token → cookie + localStorage delivery`

`InterstitialRenderer` already exposes every hook the widget needs (`extra_head`, `form_fields`, `submit_disabled`, `submit_guard`, `extra_script`), and `challenge.audience` already defaults to the provider name, so Turnstile pass tokens will not cross-verify against a `math` or `altcha` instance sharing the same secret.

## What is new

Unlike `math` (HMAC-signed answer) and `altcha` (locally re-hashable proof of work), `cf-turnstile-response` is an opaque Cloudflare artifact. There is no offline verification format, so `verifySolution()` must make a server-to-server call:

```
POST https://challenges.cloudflare.com/turnstile/v0/siteverify
  secret=<secret key>&response=<token>[&remoteip=…]
→ {"success":bool,"challenge_ts":…,"hostname":…,"error-codes":[…]}
```

This is the first outbound HTTP call in the challenge path. It follows the precedent set by `AbuseIpdb::fetch()` — stream wrapper with `stream_context_create()`, no new Composer dependency, deliberately keeping a security library's dependency surface flat (and consistent with #124 dropping Guzzle). The call lives in a `protected fetch()` so tests can stub it.

Replay protection needs nothing: Cloudflare returns `timeout-or-duplicate` for an already-validated token, so solutions are single-use at the source and `SingleUseSolutionInterface` is not required.

## Config surface

```yaml
challenge:
  provider: turnstile
  secret: '%env(FIREWALL_CHALLENGE_SECRET)%'   # HMAC key for pass tokens
  path: /_firewall/challenge
  provider_options:
    site_key: '%env(TURNSTILE_SITE_KEY)%'      # public, rendered into the widget
    secret_key: '%env(TURNSTILE_SECRET_KEY)%'  # private, sent to siteverify
    timeout: 2                                  # siteverify timeout, seconds
    on_error: block                             # block|allow when siteverify is unreachable
    send_remoteip: false                        # forward client IP to Cloudflare
    theme: auto                                 # auto|light|dark
```

Missing `site_key` or `secret_key` throws `ConfigurationException` from the constructor, so a misconfiguration is a startup failure rather than a challenge every visitor fails.

Two naming hazards the docs must call out explicitly: `challenge.secret` is the pass-token HMAC key and is **not** Turnstile's `secret_key`; and `site_key` / `secret_key` have opposite visibility despite the similar names.

## Design decisions

1. **Fail closed when `siteverify` is unreachable.** An unverifiable solution has not been verified. `on_error: allow` exists for operators who would rather absorb the risk than lock visitors out during a Cloudflare incident, but the default blocks. The cost is explicit: a Cloudflare outage makes challenged routes impassable.
2. **`remoteip` is opt-in, default off.** It improves Cloudflare's signal, but is actively harmful when `global.require_trusted_proxies` is off — a spoofed `X-Forwarded-For` would send a bogus IP and fail verification for legitimate visitors. The provider cannot see that global, so it must not infer.
3. **No shared abstract base yet.** Write Turnstile concretely; extract `AbstractRemoteCaptchaProvider` when hCaptcha lands, with two real implementations to generalise from.
4. **Turnstile does not become the default provider.** `math` and `altcha` need no reachable third party (altcha's bundle can be self-hosted). A visitor whose network blocks `challenges.cloudflare.com` can never pass a Turnstile challenge. Docs should frame it as a per-route choice.

## Known limitations to document

- **No Subresource Integrity.** Cloudflare's `api.js` is unversioned and mutable, so a pinned `integrity` digest would break on their next deploy. This is a real downgrade from `altcha`, which pins an exact version plus an SRI hash.
- **CSP requirements.** Operators need `script-src` and `frame-src` entries for `https://challenges.cloudflare.com`.
- **Latency.** One outbound round-trip per challenge submission (not per request — only on the POST to `challenge.path`).

## Scope

- [ ] `src/Challenge/TurnstileChallengeProvider.php`
- [ ] Register `turnstile` in `ChallengeProviderFactory::BUILTINS`
- [ ] `tests/Unit/Challenge/TurnstileChallengeProviderTest.php` with `fetch()` stubbed
- [ ] Integration coverage in `tests/Integration/ChallengeFlowTest.php`
- [ ] `config/config.yml` + `example/config.notes.yml` documentation
- [ ] `docs/plugins/challenges.md` built-in provider section, replacing the custom-provider sketch
- [ ] `example/demo/config.turnstile.yml` + demo route, using Cloudflare's dummy keys

sean-e-dietrich:
## Description

Adds Cloudflare Turnstile as a built-in `challenge.provider`, alongside `math` and `altcha`. Previously `docs/plugins/challenges.md` told operators who wanted real bot resistance to write their own provider and shipped a partial Turnstile sketch as the worked example, leaving server-side verification, escaping, timeouts and CSP to each of them individually.

## Related Issue

Fixes #125

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Changes Made

- **`src/Challenge/TurnstileChallengeProvider.php`** — renders the Turnstile widget through `InterstitialRenderer`, verifies the posted token against Cloudflare's siteverify API.
- **`ChallengeProviderFactory`** — registers the `turnstile` short name, and now matches constructor parameters **by declared type** rather than counting them.
- **`InterstitialRenderer`** — new optional `submit_failure` hook.
- **Tests** — 46 unit cases, 7 transport cases, 8 integration cases, factory cases for the new injection rule, and Turnstile added to the inline-script parse guard.
- **`tests/Traits/ChallengeNamespaceOverrides.php`** — namespace-scoped `fopen()` / `stream_get_meta_data()` shims for `Kanopi\Firewall\Challenge`, so the siteverify transport can be exercised without a network. Mirrors the existing `PluginsNamespaceOverrides.php`.
- **Docs / config / demo** — built-in provider section with a comparison table, `config/config.yml` and `example/config.notes.yml` option reference, and a `/secure-turnstile` demo route.

### The one architectural difference

`math` re-checks an HMAC it signed; `altcha` re-runs a hash locally. Both decide validity from the posted payload alone. A `cf-turnstile-response` token is opaque — no published format or key verifies it offline — so `verifySolution()` makes one server-to-server call on the challenge POST (never on ordinary requests). It uses the stream wrapper, matching `AbuseIpdb::fetch()`, so no new Composer dependency and the dependency surface stays flat after #124 dropped Guzzle.

### Decisions worth reviewing

| Decision | Rationale |
|---|---|
| **Fail closed** when siteverify is unreachable (`on_error: block` default) | A verdict that could not be obtained is not a pass. Failing open would make any siteverify disruption — outage, DNS failure, egress filtering — a bypass for every challenged route, on demand for anyone who can cause it. The cost is explicit in the docs: challenged routes are impassable while Cloudflare is unreachable. `on_error: allow` opts out, and only governs unreachability — a token Cloudflare actively rejects is always refused. |
| **`send_remoteip` off by default** | `getClientIp()` is only trustworthy with `global.require_trusted_proxies` configured. Behind an unconfigured proxy it returns whatever `X-Forwarded-For` claimed, so a spoofed header sends Cloudflare an unrelated address and fails verification for legitimate visitors. A provider cannot see that global, so it must not infer. |
| **`internal-error` counts as unreachability** | Cloudflare documents it as retryable, so it is not a verdict on the token. |
| **No `SingleUseSolutionInterface`** | Cloudflare answers `timeout-or-duplicate` to a replayed token, so solutions are single-use at the source. No storage record, unlike `altcha`. |
| **Missing keys throw at startup** | The alternative is a firewall serving a widget every visitor is guaranteed to fail. |
| **No abstract base class yet** | Written concretely; extract `AbstractRemoteCaptchaProvider` when hCaptcha lands, with two real implementations to generalise from. |
| **Turnstile is not the new default** | `math` and `altcha` need no reachable third party. A visitor whose network blocks `challenges.cloudflare.com` can never pass. Documented as a per-route choice. |

### Two supporting changes

**Type-based constructor injection.** Turnstile signs no state of its own, so accepting a `TokenManager` would be a lie in its signature — and PHPStan at level max rejects the unused parameter, correctly. The factory now passes `provider_options` to an `array` parameter and the `TokenManager` to anything else. Backward compatible in both directions: untyped parameters still receive the `TokenManager`, so custom providers written against the previous fixed `new $class($tokenManager)` signature are unaffected. Covered by four new factory tests including the untyped and no-constructor cases.

**`submit_failure` renderer hook.** A refused Turnstile submission has spent its token — Cloudflare answers `timeout-or-duplicate` to any further use — so without recycling the widget the visitor is left clicking **Continue** on something that can never succeed, with only a page reload as a way out. The hook is optional and defaults to empty, so `math` and `altcha` render byte-identical output.

### Unrelated fix included

`example/config.notes.yml` documented `secret: '${FIREWALL_CHALLENGE_SECRET}'`, but `TokenSubstitute` only implements `%env(VAR)%` — `${VAR}` is taken literally. Anyone who copy-pasted it got a pass-token HMAC secret that is the literal string `${FIREWALL_CHALLENGE_SECRET}`, identical across every deployment that followed the docs. Corrected in the challenge section I was already editing, since leaving it would have contradicted the Turnstile examples right above it.

**The same broken syntax appears in `docs/configuration/logging.md` (9 occurrences) and is left alone here** — it deserves its own issue rather than being buried in a feature PR. Filed as #129.

### How to Test

Automated:

```bash
composer test      # 1188 unit + 57 integration, all green
composer phpunit   # both suites in one pass, with coverage — what the CI gate runs
composer check     # PHPCS, PHPStan level max, Rector — all clean
```

Against the real Cloudflare API, using their published dummy keys:

```bash
composer demo
open http://localhost:8000/secure-turnstile
```

Verified locally end to end:

1. `/secure-turnstile` serves the interstitial — widget present, `data-sitekey` rendered, **Continue** ships `disabled`, `redirect_to=/secure-turnstile` and `ttl=60` echoed back.
2. POST to `/_firewall/challenge-turnstile` → real siteverify round trip → `{"token":"…","redirect":"/secure-turnstile"}` plus a `fw_challenge_turnstile_pass` cookie. The decoded token carries `"aud":"turnstile"`.
3. Replaying that cookie on `/secure-turnstile` passes through to the demo page.

The live transport was also exercised directly against all three dummy secrets. This is now backed by automated tests as well (see the coverage note below), but the live run is what confirms the request Cloudflare actually accepts:

| Secret key | Result |
|---|---|
| `1x0000000000000000000000000000000AA` (always passes) | `true` |
| `2x0000000000000000000000000000000AA` (always fails) | `false` |
| `3x0000000000000000000000000000000AA` (token already spent) | `false` |
| empty token (no request sent) | `false` |

### Test Configuration

```yaml
challenge:
  provider: turnstile
  secret: '%env(FIREWALL_CHALLENGE_SECRET)%'
  path: /_firewall/challenge
  cookie_name: fw_challenge_turnstile_pass
  provider_options:
    site_key: '1x00000000000000000000AA'
    secret_key: '1x0000000000000000000000000000000AA'

plugins:
  - plugin: "Kanopi\\Firewall\\Plugins\\Url"
    response: challenge
    weight: 0
    enable: true
    metadata:
      default_expiration_time: 60
    config:
      - "path@starts_with:/secure-turnstile"
```

### Code Quality

- [x] My code follows the project's code style (`composer cs` passes)
- [x] Static analysis passes (`composer stan` passes)
- [x] I have added appropriate code comments explaining complex logic
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors

### Testing

- [x] I have added unit tests that prove my fix is effective or that my feature works
- [x] I have added integration tests if necessary
- [x] New and existing tests pass locally (`composer test` passes)
- [x] I have achieved 100% test coverage for new code — **now met. See below.**
- [x] I have tested edge cases and error conditions

**Coverage, resolved.** An earlier revision of this description recorded `TurnstileChallengeProvider.php` at 83.5% (111/133), with the 22 uncovered statements being the body of `fetch()` — `stream_context_create()` / `fopen()` / `stream_get_meta_data()` / `fclose()`. **That file is now at 100% (133/133).**

The fix came from merging `2.x`, which brought in #132 and with it the pattern for this exact problem. `AbuseIpdbTransportTest` covers the identical stream-wrapper transport by defining `fopen()` and `stream_get_meta_data()` inside the plugin's own namespace, where PHP resolves unqualified calls before falling back to the global function. `AbuseIpdb.php` went from 65.8% to 100% that way. The same approach is now applied here via `tests/Traits/ChallengeNamespaceOverrides.php` — inert unless a flag is set, reset in `tearDown()`, and delegating untouched for any handle it did not create, so `math` and `altcha` are unaffected.

Seven cases in `tests/Unit/Challenge/TurnstileTransportTest.php` drive the real `fetch()` end to end: a verified token, a rejected one, an unreachable endpoint under both `on_error` settings, a non-200, a response whose headers carry no parseable status line, and the `send_remoteip` opt-in reaching the request body.

The last two are worth a reviewer's eye. A well-formed `success: true` body arriving with no readable status line must still be refused — otherwise a broken proxy becomes a free pass — and only an *absent* `wrapper_data` key reproduces it, so the shim distinguishes an absent `headers` key from an empty one.

Note that #127 is superseded by #131; making these wrappers injectable is no longer the plan, since the namespace shim reaches the same paths without changing production signatures.

**Repo-wide coverage.** The figures in the earlier revision (92.11% against an 85% gate) are stale — #132 raised the gate to 90 and took the project to 100%. Measured locally on this branch after the merge: **98.76% (3896/3945)**. The whole of the shortfall is `RedisRateLimitStorage.php` (0/49), skipped by `#[RequiresPhpExtension('redis')]` on machines without `ext-redis`; 3896 + 49 = 3945, so CI — which does `pecl install redis` — reports 100%. Every line reachable without that extension is covered.

### Security Considerations

- [x] I have considered the security implications of my changes
- [x] Input validation is properly implemented
- [x] No sensitive information is logged or exposed
- [x] Rate limiting is considered for new endpoints/features
- [x] Changes don't introduce new attack vectors

Specifics:

- **The secret key never reaches the browser.** Asserted by a test that renders the interstitial and greps for it.
- **`verifySolution()` cannot be made to throw.** `InputBag::get()` raises `BadRequestException` on a non-scalar, so posting `cf-turnstile-response[]=x` through it would turn attacker input into a 500. The token is read off `->all()` instead, with a test for the array case. *`math` and `altcha` do read through `get()` and have this defect — since confirmed end-to-end (the exception escapes `Firewall::evaluate()` uncaught) and filed as #130. Not touched here.*
- **No round trip for input that cannot be valid** — an empty or over-2048-byte token is refused unsent, so the endpoint cannot be used to make this server flood Cloudflare.
- **`success` must be a JSON `true`**, not merely truthy. Tested.
- **`error-codes` is normalised** rather than trusted, so a hostile or malformed array cannot reach the logger.
- **Configuration faults are logged at `error`** and distinguished from failed visitors, so a wrong secret key does not read as a wave of bots.
- **No SRI, deliberately** — documented, with the CSP directives operators need.

### Documentation

- [x] I have added/updated configuration examples
- [x] I have documented any new plugin variables or options
- [x] I have updated inline documentation/comments
- [ ] I have updated the README.md if needed — README does not enumerate providers, so no change was needed

## Breaking Changes

The factory's injection rule changed, but it is backward compatible: every previously valid provider signature is constructed exactly as before. Existing `math` and `altcha` interstitials render byte-identical output.

## Performance Impact

- [x] This change has performance implications

One outbound HTTPS round trip per challenge **submission** — not per request. Ordinary traffic never reaches this provider, and the interstitial itself renders with no network call. The timeout defaults to 2s (clamped 1–10) because it runs while the visitor waits. No change to any existing provider's path.

## Reviewer Notes

Most useful places to push back:

1. **`on_error: block` as the default.** This is the judgement call with real operational teeth — a Cloudflare incident makes challenged routes impassable. I think fail-closed is right for a security library and the escape hatch is one line of config, but it is worth a second opinion.
2. **The factory change.** It resolves a genuine PHPStan complaint rather than papering over it, but it does touch a contract custom providers depend on. The BC argument is that untyped and `TokenManager`-first signatures are unchanged — please check I have not missed a shape.
3. **`internal-error` mapped to unreachability.** Defensible from Cloudflare's docs, but it does mean a sustained internal error behaves like an outage under `on_error`.
4. **`submit_failure` touching shared rendering.** Small and additive, but it is the one change that affects the other two providers' code path.
5. **The namespace shim for the transport tests.** It follows the precedent set by `PluginsNamespaceOverrides.php` on `2.x` rather than inventing anything, but function shadowing is a sharp tool: the flags are process-global, so a leaked one would feed a canned HTTP response to every later test in the run. They are reset in `tearDown()` — worth confirming there is no path that skips it.

## Rebase note

`2.x` has been merged into this branch (`940cca9`), picking up #132's coverage work. That merge is what surfaced the `fetch()` gap and supplied the pattern used to close it.

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -14,8 +14,10 @@ plugins: []
 # are configured this section is ignored.
 #   secret:      HMAC key for signing pass tokens. REQUIRED when challenge
 #                plugins are active. Use a long random string, typically
-#                injected from an env var (Config supports `${VAR}`).
-#   provider:    Built-in short name (`math`) or a FQCN that implements
+#                injected from an env var (`%env(VAR)%`). Unrelated to any
+#                third-party challenge service's secret key.
+#   provider:    Built-in short name (`math`, `altcha`, `turnstile`) or a
+#                FQCN that implements
 #                Kanopi\Firewall\Challenge\ChallengeProviderInterface.
 #   path:        URL the interstitial form POSTs to.
 #   cookie_name: Cookie that carries the pass token after a successful
@@ -26,7 +28,8 @@ plugins: []
 #   provider_options:
 #                Optional per-provider settings forwarded to the provider
 #                constructor. `altcha` accepts `widget_src` and
-#                `widget_integrity`; see example/config.notes.yml.
+#                `widget_integrity`; `turnstile` REQUIRES `site_key` and
+#                `secret_key`. See example/config.notes.yml.
 #   audience:    Scopes pass tokens to this instance so two firewalls
 #                sharing a secret do not accept each other's tokens.
 #                Defaults to the provider name when left empty.
--- a/docs/guides/demo.md
+++ b/docs/guides/demo.md
@@ -1,6 +1,6 @@
 # Lite Firewall local demo
 
-A throwaway harness for poking at the firewall in a browser. Demonstrates all three response modes (`allow`, `block`, `challenge`) and both built-in challenge providers (`math` and `altcha`) side by side, and ships a production-shaped nginx → php-fpm stack so you can run perf experiments against something that looks like a real deployment.
+A throwaway harness for poking at the firewall in a browser. Demonstrates all three response modes (`allow`, `block`, `challenge`) and all three built-in challenge providers (`math`, `altcha` and `turnstile`) side by side, and ships a production-shaped nginx → php-fpm stack so you can run perf experiments against something that looks like a real deployment.
 
 ## TL;DR — composer shortcuts
 
@@ -24,6 +24,7 @@ The config keys rules off the URL so behavior is identical on any networking set
 | `/admin`          | Blocked — a `response: block` URL plugin returns 400 with the configured banning message.       |
 | `/secure`         | Challenged via the **math** provider. Solve "What is A + B?" → 60s pass cookie (`fw_challenge_pass`). |
 | `/secure-altcha`  | Challenged via the **ALTCHA** provider. The widget solves a SHA-256 proof-of-work automatically → 60s pass cookie (`fw_challenge_altcha_pass`). |
+| `/secure-turnstile` | Challenged via the **Cloudflare Turnstile** provider, using Cloudflare's always-passes [test keys](https://developers.cloudflare.com/turnstile/troubleshooting/testing/) → 60s pass cookie (`fw_challenge_turnstile_pass`). Unlike the other routes this one needs outbound network access: the widget loads from Cloudflare and the token is verified against their siteverify API. |
 
 ### What each route looks like
 
@@ -54,7 +55,7 @@ The config keys rules off the URL so behavior is identical on any networking set
     request to any route is rejected by the blocklist before plugins run. Run
     `composer demo:reset` between experiments to start clean.
 
-Only one `challenge.provider` is configurable per Firewall instance, so the demo ships two configs (`config.yml` for math, `config.altcha.yml` for ALTCHA) and `index.php` dispatches between them based on the request path. The two pass tokens live in distinct cookies and submit paths so they coexist in one browser session.
+Only one `challenge.provider` is configurable per Firewall instance, so the demo ships three configs (`config.yml` for math, `config.altcha.yml` for ALTCHA, `config.turnstile.yml` for Turnstile) and `index.php` dispatches between them based on the request path. Each pass token lives in its own cookie and submit path so they coexist in one browser session.
 
 Note that the distinct cookie names are ergonomics, not a security boundary. The two instances share a `challenge.secret`, so what actually stops a token earned on the easy math challenge from opening `/secure-altcha` is the `aud` claim, which defaults to the provider name — see [Scoping tokens across instances](../plugins/challenges.md#scoping-tokens-across-instances). Try it: solve `/secure`, copy `fw_challenge_pass` into `fw_challenge_altcha_pass`, and `/secure-altcha` still challenges you.
 
--- a/docs/plugins/challenges.md
+++ b/docs/plugins/challenges.md
@@ -14,8 +14,8 @@ The pass token is:
 
 ```yaml
 challenge:
-  provider: math                # 'math' is the built-in; or a FQCN that implements ChallengeProviderInterface
-  secret: "${FIREWALL_SECRET}"  # REQUIRED. Long random string, ideally from an env var.
+  provider: math                # 'math', 'altcha' or 'turnstile'; or a FQCN implementing ChallengeProviderInterface
+  secret: '%env(FIREWALL_CHALLENGE_SECRET)%'   # REQUIRED. Long random string, ideally from an env var.
   cookie_name: fw_challenge_pass
   header_name: X-Firewall-Challenge
   path: /_firewall/challenge    # The URL the interstitial POSTs to
@@ -57,7 +57,7 @@ If you run **the same provider** in several places with the same secret — say 
 ```yaml
 challenge:
   provider: altcha
-  secret: "${FIREWALL_SECRET}"
+  secret: '%env(FIREWALL_CHALLENGE_SECRET)%'
   audience: admin-portal   # defaults to the provider name
 ```
 
@@ -67,7 +67,7 @@ The alternative is to give each instance its own `challenge.secret`, which isola
 
 ## Built-in providers
 
-Two providers ship with the firewall — set `challenge.provider` to either short name:
+Three providers ship with the firewall — set `challenge.provider` to a short name:
 
 === "math"
 
@@ -84,8 +84,24 @@ Two providers ship with the firewall — set `challenge.provider` to either shor
     **Continue** on its own. The visitor clicks nothing; a bot pays CPU for
     every solve.
 
-Both screenshots come from the [demo application](../guides/demo.md), which
-serves each provider on its own route.
+=== "turnstile"
+
+    Cloudflare's Turnstile widget, verified server-side against their
+    siteverify API. The strongest bot resistance of the three, and the only
+    one that depends on a third party being reachable — from the visitor's
+    browser *and* from your server.
+
+The math and ALTCHA screenshots come from the [demo application](../guides/demo.md), which serves each provider on its own route.
+
+Which to pick is mostly a question of what you are willing to depend on:
+
+| | `math` | `altcha` | `turnstile` |
+|---|---|---|---|
+| Third-party script | none | CDN (self-hostable) | Cloudflare only |
+| Outbound call from your server | none | none | one per submission |
+| Subresource Integrity | n/a | yes, pinned | not possible |
+| Bot resistance | lowest | CPU cost per solve | highest |
+| Fails if the third party is down | n/a | interstitial degrades | nobody can pass (default) |
 
 - **`math`** — asks "What is A + B?" with single-digit operands. Low-friction proof-of-effort, no JS bundle, no external script load. Defeats the laziest bots; trivial for a human.
 - **`altcha`** — embeds the [ALTCHA](https://altcha.org/docs/v2/) v2 widget with a pre-computed challenge (no server round-trip to fetch one). The visitor's browser brute-forces `SHA-256(salt + N) == challenge`; the salt embeds an expiry and the challenge is HMAC-signed with `challenge.secret`, so the server stays stateless. Privacy-respecting, and imposes a per-solve CPU cost on bots. Solved challenges are single-use — see [Single-use solutions](#single-use-solutions).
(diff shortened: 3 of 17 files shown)


kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 93.36
Passing: 85

[Report](https://output.circle-artifacts.com/output/job/a61d0801-1983-452c-9642-56abbd6a5d48/artifacts/0/reports/index.html)

kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/c74211d3-1fef-45b6-86ea-b7957b61050f/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/c74211d3-1fef-45b6-86ea-b7957b61050f/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/6eea24a2-8cd6-4968-a49a-96fa2750b4a4/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/7f7ff869-7906-4ff5-856f-aa805d13fbe4/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/779588a3-89b7-470a-90bb-adff7393ff7c/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/a2a3411c-a707-4fb8-aa09-9ab74e1d970d/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/4aededdf-6c9b-4642-9cbe-6996e6b183cc/artifacts/0/reports/index.html)

Issue #131, opened 2026-08-04
Measure coverage across both suites, then work toward 100% (supersedes #127)


sean-e-dietrich:
## Summary

Coverage is measured across the unit suite only, because `composer phpunit:integration` runs with `--no-coverage`. Fixing that one thing moves lines from **92.40% → 94.61%** and `src/Firewall.php` from **73.59% → 94.1%** without writing a single test. This issue supersedes #127, whose diagnosis was wrong (see below), and lays out the actual route to 100%.

## Corrects #127

#127 claims `Firewall.php`'s 103-statement gap is "behind `@codeCoverageIgnoreStart` or the `exit()`-terminated production response path", and proposes a redesign to reach it. **That is incorrect.** PHPUnit excludes `@codeCoverageIgnore`-annotated lines from the denominator entirely, so those paths were never counted against coverage in the first place. Confirmed by inspection: the ignored block at `Firewall.php:696-703` does not appear in the uncovered-line set at all.

The 103 uncovered statements were ordinary, reachable code — `handleChallengeSubmission()` (668-718), `consumeSingleUseSolution()` (750-773), `sendChallengeResponse()` (786-814), `createChallengePieces()` (283-301) — all of which `tests/Integration/ChallengeFlowTest.php` exercises heavily. They read as uncovered purely because integration tests are not measured.

No redesign of `Firewall.php` is needed. Please close #127 in favour of this issue.

## Measured state, at `03bc10f` on `2.x`

| Metric | Unit only (current) | Both suites | Delta |
|---|---:|---:|---:|
| Lines | 92.40% (3514/3803) | **94.61% (3598/3803)** | +84 lines |
| Methods | 85.34% (262/307) | **87.62% (269/307)** | +7 methods |
| Classes | 57.45% (27/47) | **61.70% (29/47)** | +2 classes |

**Read the class figure, not the line figure.** A class counts only when every line in it is hit, so it cannot be flattered by a mostly-covered file hiding a dead branch. 94.61% of lines sounds nearly done; 61.70% of classes is the honest measure. The CI gate at `.circleci/config.yml:143` reads the **line** percentage — the most forgiving of the three.

## Remaining gaps once both suites are measured

19 files below 100%. Grouped by what each actually needs:

### Already tested, just not in this environment (72 statements)

| File | Missing | Blocker |
|---|---:|---|
[...]

Issue #128, opened 2026-08-04
Drop unused CI setup dependencies and the flaky bc/xidel downloads


sean-e-dietrich:
## Summary

The `phpunit` job's `Setup` step installs four things it does not need, two of them by fetching binaries from third-party hosts mid-build. One of those fetches broke a build on #126 for reasons unrelated to the change under test. All four can go, and the work they do collapses into one `php -r` call using the interpreter already in the image.

## What broke

`ci/circleci: phpunit 8.2` on #126 failed in `Setup`, before PHPUnit ran:

```
curl: (56) Recv failure: Connection reset by peer
Exited with code exit status 56
```

That is [`.circleci/config.yml:105`](../blob/2.x/.circleci/config.yml#L105):

```bash
curl -o /tmp/bc.deb -fsSL http://archive.ubuntu.com/ubuntu/pool/main/b/bc/bc_1.07.1-3build1_amd64.deb
```

A single version-pinned `.deb`, over **plain HTTP**, from `archive.ubuntu.com`, on every run of all five PHP versions. The same commit passed `phpunit` on 8.1, 8.3, 8.4 and 8.5, so this was pure transport flakiness. A rerun went green with no code change.

## Audit of the Setup step

| Installed | What actually uses it | Verdict |
|---|---|---|
| `mariadb-client`, `postgresql-client` (apt) | Nothing. No `mysql`, `mysqladmin`, `psql` or `pg_isready` invocation exists anywhere in `.circleci/`, `bin/` or `tests/`. Database tests connect through the `pdo_mysql` / `pdo_pgsql` extensions installed two lines below. | **Remove** |
| `bc` (curl'd `.deb`, plain HTTP) | Exactly one line — `OUTCOME=$(echo "${PERCENTAGE} < ${PASSING}" \| bc)` at `config.yml:145` | **Remove** |
| `xidel` (curl'd GitHub release tarball) | Exactly one line — the XPath at `config.yml:143` pulling a coverage percentage out of `reports/xml/index.xml` | **Remove** |
| `pecl install xdebug` | Coverage | Keep |
| `pecl install redis` | `RedisRateLimitStorage` tests | Keep |
| `docker-php-ext-install pdo_mysql mysqli pdo_pgsql` | `DatabaseStorage` tests | Keep |

So the coverage gate reaches out to two third-party hosts to install two binaries, in order to perform one arithmetic comparison and one XPath query — inside a container that ships PHP.

## Proposed change

[...]

Issue #127, opened 2026-08-04
Close the test coverage gap and settle the real coverage target


sean-e-dietrich:
## Summary

`CONTRIBUTING.md` and the PR template both require **100% test coverage for new code**, but the CI gate is set to 85% and actual line coverage is 92.11%. 21 of 58 source files are below 100%. This issue records where the gaps actually are, and separates the ones that are real from the ones that are measurement artifacts.

## Measured state

Unit suite, `XDEBUG_MODE=coverage`, at `8dbc0ca`:

```
statements: 3633/3944 = 92.11%
methods:     273/319  = 85.58%
21 of 58 files below 100%
```

| File | Statements | Covered | Missing | % |
|---|---:|---:|---:|---:|
| `src/Firewall.php` | 390 | 287 | **103** | 73.6% |
| `src/Plugins/AbuseIpdb.php` | 202 | 133 | **69** | 65.8% |
| `src/RateLimitStorage/RedisRateLimitStorage.php` | 49 | 0 | **49** | 0.0% |
| `src/Storage/DatabaseStorage.php` | 304 | 281 | 23 | 92.4% |
| `src/Challenge/TurnstileChallengeProvider.php` | 133 | 111 | 22 | 83.5% |
| `src/Plugins/UserAgent.php` | 174 | 161 | 13 | 92.5% |
| `src/Challenge/AltchaChallengeProvider.php` | 92 | 87 | 5 | 94.6% |
| `src/Plugins/Crs.php` | 109 | 104 | 5 | 95.4% |
| `src/Utility/TokenSubstitute.php` | 318 | 314 | 4 | 98.7% |
| `src/Challenge/TokenManager.php` | 42 | 39 | 3 | 92.9% |
| `src/Exception/ChallengeSolvedException.php` | 3 | 0 | 3 | 0.0% |
| `src/Logging/LoggingTrait.php` | 30 | 28 | 2 | 93.3% |
| `src/Traits/AddressMatchTrait.php` | 25 | 23 | 2 | 92.0% |
| `src/Challenge/MathChallengeProvider.php` | 48 | 47 | 1 | 97.9% |
| `src/Exception/ChallengeRequiredException.php` | 1 | 0 | 1 | 0.0% |
| `src/Storage/InMemoryStorage.php` | 125 | 124 | 1 | 99.2% |
| `src/Traits/FileTrait.php` | 94 | 93 | 1 | 98.9% |
[...]

PR #132, opened 2026-08-04
Measure coverage correctly, enforce it, and take it to 100%

sean-e-dietrich:
## Description

Coverage was measured on the unit suite alone, the CI gate that reported it had been commented out, and three tests were passing while asserting nothing about the behaviour they named. This fixes the measurement, turns the gate on, and closes every remaining gap.

**Line, method and class coverage are all now 100% on CI**, up from a real 92.40%.

## Related Issue

Refs #131

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Refactoring (no functional changes)

## Results

Measured on this branch:

| Metric | Before (unit only) | After (both suites, CI) |
|---|---:|---:|
| Lines | 92.40% (3514/3803) | **100.00% (3804/3804)** |
| Methods | 85.34% (262/307) | **100.00% (308/308)** |
| Classes | 57.45% (27/47) | **100.00% (47/47)** |

## Read this first: the starting number was wrong

Roughly a third of the gain was a measurement fault, not missing tests.

`composer phpunit:integration` ran with `--no-coverage`, so 57 integration tests covering 474 assertions contributed nothing to the figure. `src/Firewall.php` read 73.59% while `tests/Integration/ChallengeFlowTest.php` was exercising `handleChallengeSubmission()`, `consumeSingleUseSolution()` and `sendChallengeResponse()` in depth. Measuring both suites in one pass moved it to 94.1% with no new tests.

Two related corrections:

- **#127's diagnosis was wrong** and #131 supersedes it. It claimed `Firewall.php`'s gap was `exit()`-terminated paths behind `@codeCoverageIgnore` and proposed a redesign to reach them. PHPUnit *excludes* annotated lines from the denominator, so those were never counted — the annotated block at `Firewall.php:696-703` does not appear in the uncovered set at all. No redesign was needed. **#127 should be closed.**
- **The gate had never run.** `PASSING=85` sat above a commented-out `if`, so nothing has ever enforced coverage. This is the first coverage figure the project has compared against anything.

## Three tests that were asserting nothing

Worth calling out separately, because these are defects the coverage work surfaced rather than coverage being an end in itself. Two were only found by asking why a line was not hit.

1. **`testDisabledModeSkipsEvaluation`** returned at the CLI bypass — `evaluate()` short-circuits under `PHP_SAPI === 'cli'` for every mode but `exception` — and never reached the disabled branch. Its assertion (no plugin manager called) is true either way. Same for log mode; `FirewallModeTest`'s own docblock had already recorded the problem for one case.
2. **`testExistingFileWithLoosePermsIsTightened`** probes for a filesystem that honours `chmod()` and skips otherwise. The CircleCI `cimg/php` images silently no-op `chmod()` on `/tmp` files — the call returns `TRUE` while `fileperms()` still shows the old mode — so on CI it never ran.
3. **A `DatabaseStorage` mock I wrote in this PR** left `where()` unstubbed, so PHPUnit auto-returned a *different* QueryBuilder mock and the assertions applied to an object the code never touched. It passed while exercising nothing. Caught because a companion assertion failed; the helper now stubs the whole fluent chain and says why.

And one blind spot this did **not** fix: `InterstitialRendererTest` shells out to `node --check`, and `node` is not on the CI image, so **24 cases silently skip there**. That guard matters — the ALTCHA and Turnstile submit buttons ship `disabled` and are only enabled by that inline script, so a syntax error locks visitors out with no fallback. It passes locally, which is how it stayed invisible. Same failure mode as the three above; worth its own issue.

## Changes to `src/` — the whole surface

Four changes across three files, all behaviour-neutral. Listed exhaustively because "source changed to make tests pass" deserves scrutiny:

| File | Change | Why |
|---|---|---|
| `src/Firewall.php` | CLI check extracted to `protected shouldBypassForCli()` | Pure extraction, identical condition. The only way to reach the `log` and `disabled` branches from a CLI test process. |
| `src/Utility/Path.php` | `\realpath()` → `realpath()` | Makes the call namespace-resolvable so the suite can shadow it. PHP falls back to the global function, so behaviour is unchanged. |
| `src/Utility/TokenSubstitute.php` | two `\file_get_contents()` → unqualified | Same reason, for the read-failed-after-`is_file()`-passed guards. |

Each carries a comment explaining why it is unqualified, so none looks like an oversight to re-qualify later. `src/Utility/Config.php` already called `is_file()` unqualified and needed no change.

Everything else is `phpunit.xml`, `composer.json`, `.circleci/config.yml` and `tests/`.

## How the hard branches were reached

Several guards cannot be triggered on demand: a read failing after `is_file()` **and** `is_readable()` both passed, `realpath()` failing on a path that exists, an HTTP API returning a specific error, `mkdir()` running for a directory a previous test run already created.

The repo already had the answer — `tests/Traits/NamespaceOverrides.php` shadows functions in `Kanopi\Firewall\Traits`. This adds the same for `Kanopi\Firewall\Utility` and `Kanopi\Firewall\Plugins`. Each shim is inert unless its `$GLOBALS` flag is set and delegates otherwise, so loading one changes nothing.

For `AbuseIpdb::fetch()` — 69 lines, the single largest gap — `fopen()` returns an in-memory stream holding a canned body and `stream_get_meta_data()` reports canned `wrapper_data`, for exactly the handles the shim created. Reaching those branches against the real API would spend quota, need a key in CI, and require AbuseIPDB to return specific errors on demand.

**One trap worth knowing.** My first shim declared `file_get_contents($filename)` and dropped everything after it. `Config::fileGetContents()` passes a stream context carrying its request timeout — silently lost, so a *different* test failed with no visible connection to the cause. Shims must forward variadically. Mine do, and the three pre-existing ones with the same latent bug (`file_put_contents`, `fopen`, `fwrite`) are fixed too.

Also used: `#[RunInSeparateProcess]` for `define()`-based branches, protected-method seams, and real config over doubles where possible — the CRS truncation test lifts the anomaly threshold rather than stubbing the engine, because CRS's own argument-count rule would otherwise block and return before the branch under test.

## Behaviours pinned along the way

Coverage is the occasion; these are the point. A sample:

- **`sanitizeRedirect()` had no tests at all** — the open-redirect guard on the challenge flow. `redirect_to` arrives on the interstitial's POST, so an attacker chooses it, and it is handed to `window.location.href` after a successful solve. Thirteen cases now, including protocol-relative `//evil.test`, which looks like a path and passes a naive "starts with `/`" check. The guard was correct; it was unverified.
- **Two opposite failure defaults, both deliberate.** AbuseIPDB fails *open* — a reputation lookup that cannot reach its API has no opinion, and turning a DNS blip into a site-wide block is worse than missing bad traffic. Turnstile (#126) fails *closed* — an unverifiable token simply has not been verified. Each now has a test naming the reasoning.
- **Correctly-signed rubbish is still rejected.** `TokenManager` and `MathChallengeProvider` verify the signature *before* parsing the payload, so the shape checks afterwards needed payloads signed with the real secret to reach them.
- **A whitelisted address never matches whatever its score** — AbuseIPDB whitelists infrastructure that attracts reports without being malicious, and blocking on score alone is how a firewall de-indexes a site.
- **The permission mask strips file-type bits.** `& 07777` before `chmod()`, because `fileperms()` returns `S_IFREG` too and some platforms — the CI image among them — refuse a chmod carrying those bits, so the tightening silently did nothing.

## Testing

```bash
composer phpunit:coverage   # both suites, one pass, with coverage
cat reports/coverage.txt
```

1171 tests pass locally (2 pre-existing deprecations, 12 skips — unchanged from `2.x`). `composer check` passes: PHPCS, PHPStan at level max, Rector.

Local coverage reads 98.71% because `RedisRateLimitStorage` needs the `redis` extension and a server; CI has both and reports it at 100%.

The gate was verified in four states, including reproducing the stdout-warning condition with `-d extension=json`:

```
94.61 >= 90.00, no warning        -> exit 0
94.61 >= 90.00, warning on stdout -> exit 0
94.61 <  99.00                    -> exit 1
report file missing               -> exit 1 with a distinct message
```

The final `FileTrait` test was verified to cover its line **in isolation**, not merely alongside the test that skips on CI — that mattered specifically because the whole problem was a test that appeared to cover a line and did not.

## CI changes

**The gate is now enforced**, at a floor of 90. That was set before the number was known and is now far below it — see the recommendation below.

**`xidel` and `bc` are gone.** Both were installed by curl'ing binaries mid-build, `bc` as a version-pinned `.deb` over **plain HTTP** from `archive.ubuntu.com`, which failed a build on #126 with `curl: (56) Recv failure: Connection reset by peer`. Making pass/fail depend on that would be worse than having no gate. Reading, reporting and comparing now happen inside one PHP process, so no shell variable holds a value that PHP's startup warnings can corrupt — this image declares three extensions twice and prints "already loaded" to stdout, which is what produced a bogus `0.00%` mid-review.

Removing the plain-HTTP `.deb` fetch is also a small supply-chain improvement: it was the one place a network-position attacker could influence the build, and its result was `dpkg -i`'d as root.

The two now-unused apt packages (`mariadb-client`, `postgresql-client`) are left to #128, which I narrowed accordingly. `rsync` in the `build` job is genuinely used and out of scope for both.

### Code Quality

- [x] `composer cs` passes
- [x] `composer stan` passes
- [x] Comments explain the non-obvious parts, particularly why three source calls are unqualified
- [x] No new warnings or errors

### Testing

- [x] Unit tests added
- [x] Integration coverage where relevant
- [x] `composer test` passes locally
- [x] 100% coverage
- [x] Edge cases and error conditions — that is most of this PR

### Security Considerations

- [x] Security implications considered
- [x] No sensitive information logged or exposed

The security-relevant additions are the `sanitizeRedirect()` cases, the correctly-signed-rubbish cases for `TokenManager`, the permission-mask test, and pinning the two fail-open/fail-closed defaults. The supply-chain note above applies too.

### Documentation

- [ ] `docs/contributing/testing.md:3` says "All new features must have 100% test coverage" while the gate is at 90

That contradiction predates this PR — the gate said 85 and enforced nothing — and settling it is the open question in #131. Now that 100% is real, the docs and the gate can finally agree; I have left them alone rather than encode a number nobody has agreed to.

## Breaking Changes

`composer test` behaves as before. `phpunit:unit` / `phpunit:integration` keep their scope and no longer emit coverage — `phpunit:coverage` is where coverage comes from.

One real change for contributors: **CI can now fail on coverage**, where previously it could not.

## Performance Impact

- [x] This change has performance implications

The CI `phpunit` step now instruments the integration suite, so it is slower — partly offset by dropping two binary downloads. `composer test` is *faster* for contributors, since it no longer runs coverage at all. No runtime change: the three source edits are a method extraction and three function calls losing a leading backslash.

## Reviewer Notes

1. **The floor should be 99, not 100.** `RedisRateLimitStorage` depends on a service container, so a floor of exactly 100 turns any infrastructure hiccup into a red build — which this PR's own history demonstrates twice over. 99 locks in essentially all of the gain and still fails on a real regression.
2. **The three `src/` edits are the thing to scrutinise.** They are listed exhaustively above. Two are function calls losing a leading backslash so tests can shadow them, which is a real (if small) trade: it makes those call sites interceptable. The alternative was leaving three defensive guards permanently unreachable or annotating them away.
3. **`@codeCoverageIgnore` was not used anywhere.** Every line is covered by a test that runs, not by an annotation. That was the failure mode worth avoiding, and the reason 100% here means something.
4. **9 commits, cleanly separable** — the first three are the measurement fix and its two follow-ups, the remaining six are coverage batches. Happy to split into two PRs if you would rather review them apart.
5. **CI is green — 28/28 — and the 100% figure reproduces across runs.** Noted only because it took a rerun to get there: three jobs had failed at `Checkout code` with `git@github.com: Permission denied (publickey)`, an SSH/deploy-key failure before anything ran. That is the second infrastructure flake this branch hit, which is the evidence behind recommendation 1 above.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,30 +147,41 @@ jobs:
             # peer". A gate that can fail for reasons unrelated to the code
             # under test is worse than no gate, and PHP is definitionally
             # present in a PHP image.
-            PERCENTAGE=$(php -r '
-              $xml = simplexml_load_file("reports/xml/index.xml");
-              echo (string) $xml->project->directory->totals->lines["percent"];
-            ')
-            # Ratchet, not a target. Set just below the figure measured when
+            # Ratchet, not a target. Set below the figure measured when
             # coverage started including the integration suite, so the gain
             # cannot silently regress. Raise it as coverage improves; see the
             # tracking issue for where the remaining gaps are.
             PASSING=90
-            echo "# PHPUnit Code Coverage" | tee -a /tmp/report.md
-            echo "PHP Version: <<parameters.php>>" | tee -a /tmp/report.md
-            echo "Percent: ${PERCENTAGE}" | tee -a /tmp/report.md
-            echo "Passing: ${PASSING}" | tee -a /tmp/report.md
-            echo "" | tee -a /tmp/report.md
-            echo "[Report](https://output.circle-artifacts.com/output/job/${CIRCLE_WORKFLOW_JOB_ID}/artifacts/0/reports/index.html)" | tee -a /tmp/report.md
-            php -r '
-              $pct = (float) $argv[1];
-              $min = (float) $argv[2];
+            # Read, report and compare entirely inside PHP. Nothing parses
+            # PHP's stdout in the shell, because on this image every `php`
+            # invocation prints "Module ... is already loaded" warnings to
+            # stdout — captured into a variable they prefix the value, and
+            # `(float) "Warning: ..."` is 0.0, which failed the gate against
+            # a real 95.90%. display_errors=stderr keeps them out of the way
+            # as well, but the comparison no longer depends on that.
+            php -d display_errors=stderr -r '
+              $min = (float) $argv[1];
+              $xml = simplexml_load_file("reports/xml/index.xml");
+              if ($xml === false) {
+                fwrite(STDERR, "Could not read reports/xml/index.xml — did the coverage run produce a report?\n");
+                exit(1);
+              }
+              $pct = (float) $xml->project->directory->totals->lines["percent"];
+              $report = sprintf(
+                "# PHPUnit Code Coverage\nPHP Version: %s\nPercent: %.2f\nPassing: %.2f\n\n[Report](https://output.circle-artifacts.com/output/job/%s/artifacts/0/reports/index.html)\n",
+                $argv[2],
+                $pct,
+                $min,
+                $argv[3]
+              );
+              file_put_contents("/tmp/report.md", $report, FILE_APPEND);
+              echo $report;
               if ($pct < $min) {
                 printf("Line coverage %.2f%% is below the %.2f%% floor.\n", $pct, $min);
                 exit(1);
               }
               printf("Line coverage %.2f%% meets the %.2f%% floor.\n", $pct, $min);
-            ' "${PERCENTAGE}" "${PASSING}"
+            ' "${PASSING}" "<<parameters.php>>" "${CIRCLE_WORKFLOW_JOB_ID}"
       - ci-tools/post-to-github-commit:
           body: '@/tmp/report.md'
           when: always
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,21 +138,39 @@ jobs:
           name: Run PHPUnit
           command: composer -n phpunit
       - run:
-          name: Check Overage
+          name: Check Coverage
           command: |
-            PERCENTAGE=$(xidel -s --xpath "string(//phpunit/project/directory/totals/lines/@percent)" reports/xml/index.xml)
-            PASSING=85
-            OUTCOME=$(echo "${PERCENTAGE} < ${PASSING}" | bc)
+            # Read and compare with PHP rather than xidel + bc. Both were
+            # installed by curl'ing binaries mid-build — one of them a
+            # version-pinned .deb over plain HTTP from archive.ubuntu.com,
+            # which has already failed a build with "Connection reset by
+            # peer". A gate that can fail for reasons unrelated to the code
+            # under test is worse than no gate, and PHP is definitionally
+            # present in a PHP image.
+            PERCENTAGE=$(php -r '
+              $xml = simplexml_load_file("reports/xml/index.xml");
+              echo (string) $xml->project->directory->totals->lines["percent"];
+            ')
+            # Ratchet, not a target. Set just below the figure measured when
+            # coverage started including the integration suite, so the gain
+            # cannot silently regress. Raise it as coverage improves; see the
+            # tracking issue for where the remaining gaps are.
+            PASSING=90
             echo "# PHPUnit Code Coverage" | tee -a /tmp/report.md
             echo "PHP Version: <<parameters.php>>" | tee -a /tmp/report.md
             echo "Percent: ${PERCENTAGE}" | tee -a /tmp/report.md
             echo "Passing: ${PASSING}" | tee -a /tmp/report.md
             echo "" | tee -a /tmp/report.md
             echo "[Report](https://output.circle-artifacts.com/output/job/${CIRCLE_WORKFLOW_JOB_ID}/artifacts/0/reports/index.html)" | tee -a /tmp/report.md
-#            if [[ "${OUTCOME}" == "1" ]]; then
-#              echo "Percentage does not pass threshold."
-#              exit 1
-#            fi
+            php -r '
+              $pct = (float) $argv[1];
+              $min = (float) $argv[2];
+              if ($pct < $min) {
+                printf("Line coverage %.2f%% is below the %.2f%% floor.\n", $pct, $min);
+                exit(1);
+              }
+              printf("Line coverage %.2f%% meets the %.2f%% floor.\n", $pct, $min);
+            ' "${PERCENTAGE}" "${PASSING}"
       - ci-tools/post-to-github-commit:
           body: '@/tmp/report.md'
           when: always
--- a/composer.json
+++ b/composer.json
@@ -69,11 +69,9 @@
         "check:code:circleci": "vendor/bin/phpcs --standard=phpcs_ruleset.xml --report=junit --report-file=phpcs-report.xml src",
         "check:code": "vendor/bin/phpcs --standard=phpcs_ruleset.xml src",
         "fix:code": "vendor/bin/phpcbf --standard=phpcs_ruleset.xml src",
-        "phpunit": [
-            "@phpunit:unit",
-            "@phpunit:integration"
-        ],
-        "phpunit:unit": "XDEBUG_MODE=coverage vendor/bin/phpunit -c ./phpunit.xml --testsuite unit --do-not-cache-result",
+        "phpunit": "@phpunit:coverage",
+        "phpunit:coverage": "XDEBUG_MODE=coverage vendor/bin/phpunit -c ./phpunit.xml --do-not-cache-result",
+        "phpunit:unit": "vendor/bin/phpunit -c ./phpunit.xml --testsuite unit --do-not-cache-result --no-coverage",
         "phpunit:integration": "vendor/bin/phpunit -c ./phpunit.xml --testsuite integration --do-not-cache-result --no-coverage",
         "fix:php": "find config src tests -name '*.php' -exec php -l {} \\;",
         "check:security": "vendor/bin/phpstan analyse src --level max --memory-limit=-1",
@@ -95,6 +93,7 @@
             "@phpunit:unit",
             "@phpunit:integration"
         ],
+        "test:gate": "@phpunit:coverage",
         "test:coverage": "XDEBUG_MODE=coverage vendor/bin/phpunit -c ./phpunit.xml --coverage-html reports/ --coverage-text",
         "cs": "@check:code",
         "cs:fix": "@fix:code",
@@ -129,6 +128,9 @@
         "check:request": "php bin/firewall-check"
     },
     "scripts-descriptions": {
+        "test": "Run both suites without coverage — the fast loop for local development",
+        "phpunit:coverage": "Run both suites in a single pass with coverage. Coverage MUST be measured across both suites: running them separately leaves integration tests out of the figure entirely, which understated Firewall.php by 20 percentage points",
+        "test:gate": "Alias for phpunit:coverage — the run the CI coverage gate reads",
         "demo": "Run the demo on PHP's built-in server (single-process) at http://localhost:8000",
         "demo:workers": "Same as `demo` but with PHP_CLI_SERVER_WORKERS=8 for concurrent request handling",
         "demo:reset": "Delete the demo's FileStorage data file (/tmp/firewall-demo.data) so the next request starts with a clean slate",
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -27,6 +27,13 @@
       <bootstrap class="DG\BypassFinals\PHPUnitExtension"/>
   </extensions>
   <php>
+    <!--
+      Covering both suites in one run needs more than the default 128M:
+      report generation holds the whole coverage set in memory and dies
+      partway through writing the HTML with "Allowed memory size exhausted",
+      which looks like a test failure but is not one.
+    -->
+    <ini name="memory_limit" value="512M" />
     <env name="FIREWALL_BYPASS_CLI" value="1" />
     <env name="FIREWALL_TEST" value="1" />
   </php>
--- a/src/Firewall.php
+++ b/src/Firewall.php
@@ -547,9 +547,8 @@ final class Firewall
      */
     public function evaluate(?Request $request = null): bool
     {
-        // Skip in CLI (Drush, cron, WP-CLI) unless mode is 'exception' (PHPUnit/framework use).
         // @codeCoverageIgnoreStart
-        if (PHP_SAPI === 'cli' && $this->firewallMode !== FirewallMode::Exception) {
+        if ($this->shouldBypassForCli()) {
             $this->getLogger()->debug('CLI mode detected, bypassing firewall');
             return true;
         }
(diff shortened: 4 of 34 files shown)


kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/c78c40d0-94fc-4204-b40a-719d2b3929cc/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/c78c40d0-94fc-4204-b40a-719d2b3929cc/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 
Warning: Module "pdo_mysql" is already loaded in Unknown on line 0

Warning: Module "mysqli" is already loaded in Unknown on line 0

Warning: Module "pdo_pgsql" is already loaded in Unknown on line 0
95.90
Passing: 90

[Report](https://output.circle-artifacts.com/output/job/411d7949-ec6a-414c-833f-a5a715bf90d1/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 
Warning: Module "pdo_mysql" is already loaded in Unknown on line 0

Warning: Module "mysqli" is already loaded in Unknown on line 0

Warning: Module "pdo_pgsql" is already loaded in Unknown on line 0
95.92
Passing: 90

[Report](https://output.circle-artifacts.com/output/job/2924faca-42e6-48de-9f98-56f73160c2d0/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 
Warning: Module "pdo_mysql" is already loaded in Unknown on line 0

Warning: Module "mysqli" is already loaded in Unknown on line 0

Warning: Module "pdo_pgsql" is already loaded in Unknown on line 0
95.92
Passing: 90

[Report](https://output.circle-artifacts.com/output/job/636dd4f1-ee29-4e0d-b459-88d9d2dc53e0/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 
Warning: Module "pdo_mysql" is already loaded in Unknown on line 0

Warning: Module "mysqli" is already loaded in Unknown on line 0

Warning: Module "pdo_pgsql" is already loaded in Unknown on line 0
95.90
Passing: 90

[Report](https://output.circle-artifacts.com/output/job/c9c4339b-6a33-4fde-8e34-7799d4cb28cd/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 
Warning: Module "pdo_mysql" is already loaded in Unknown on line 0

Warning: Module "mysqli" is already loaded in Unknown on line 0

Warning: Module "pdo_pgsql" is already loaded in Unknown on line 0
95.92
Passing: 90

[Report](https://output.circle-artifacts.com/output/job/a860fe55-d1c1-4ce3-8e45-9749dbb7eba8/artifacts/0/reports/index.html)

kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/fa47536e-6104-4e53-957a-ca6fe0f9183a/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/fa47536e-6104-4e53-957a-ca6fe0f9183a/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 95.92
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/45bbc04d-4bec-4b50-ac96-8f2b4175e6a4/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 95.92
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/1a0a2639-8295-46f0-aa0b-2be55b2a1743/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 95.90
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/087126f6-6b70-4b5b-b728-00b286485105/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 95.90
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/e9327717-fe1a-41a4-8aea-4502a6e5a8f7/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 95.92
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/50dca172-3341-412c-b3a2-a1f3bd1d94d2/artifacts/0/reports/index.html)

kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/64c49a30-fecc-41ea-a64d-41e6c7034a51/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/64c49a30-fecc-41ea-a64d-41e6c7034a51/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 97.58
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/94d8fd61-d684-4ff9-b4ed-4038911cdcdf/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 97.56
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/8f8f79f5-1067-427f-b87a-58c0f6bf0da9/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 97.56
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/2c132b0c-151f-4375-8a07-0c1ac8fd2b17/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 97.58
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/8ceac9bb-c761-459f-b15e-ba06f8a36801/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 97.58
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/488043d7-7016-4051-a62a-eef78f1b2e8f/artifacts/0/reports/index.html)

kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/66b034cb-0526-49e4-936f-d747d204f672/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/66b034cb-0526-49e4-936f-d747d204f672/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 99.97
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/c5ccb982-839a-4faa-95b4-927013475139/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 99.97
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/170b13e4-edeb-4c9c-b0b0-08f411f84b96/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/7c2f94df-d9ed-4c3b-be29-c86e3dd79bac/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/57cfcbb6-3904-4cfb-b62b-e6946fa10b89/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/642ae51b-f502-4210-ba95-d3785d08f94c/artifacts/0/reports/index.html)

kanopicode:
📖 **Docs preview** for this PR: [https://output.circle-artifacts.com/output/job/86e6b999-51fc-458a-8ece-51cf574dea96/artifacts/0/docs/index.html](https://output.circle-artifacts.com/output/job/86e6b999-51fc-458a-8ece-51cf574dea96/artifacts/0/docs/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.3
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/b6a4b241-f3ba-470a-940b-1a978e61b12e/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.5
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/5fe61277-42e3-4b85-a280-38a957c0b849/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.2
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/9d2bd20e-a305-4607-9b1d-83bd458198d5/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.4
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/1f52c03c-4011-4e71-b744-191a5eabc72c/artifacts/0/reports/index.html)

kanopicode:
# PHPUnit Code Coverage
PHP Version: 8.1
Percent: 100.00
Passing: 90.00

[Report](https://output.circle-artifacts.com/output/job/532f74bd-b487-4bbc-8fd4-c36ef4d02b6d/artifacts/0/reports/index.html)

